PR #3, opened 2019-10-31
Tokens Module

shaopengw:
Version 1 impl for tokens module #4

### DIFF
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -8,7 +8,7 @@ use sr_primitives::traits::{
 };
 // FIXME: `srml-` prefix should be used for all srml modules, but currently `srml_system`
 // would cause compiling error in `decl_module!` and `construct_runtime!`
-// #3296 https://github.com/paritytech/substrate/issues/3295
+// #3295 https://github.com/paritytech/substrate/issues/3295
 use srml_system::{self as system, ensure_signed};
 
 use traits::MultiCurrency;
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -1,21 +1,43 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use support::{decl_event, decl_module, decl_storage};
+use codec::Codec;
+use rstd::fmt::Debug;
+use srml_support::{decl_event, decl_module, decl_storage, Parameter, StorageMap};
+use sr_primitives::traits::{MaybeSerializeDeserialize, Member, SimpleArithmetic};
+//FIXME: `srml-` prefix should be used for all srml modules, but currently `srml_system`
+// would cause compiling error in `decl_module!` and `construct_runtime!`
+// #3296 https://github.com/paritytech/substrate/issues/3295
+use srml_system as system;
+
 use traits::MultiCurrency;
 
-pub trait Trait: system::Trait {
-	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+pub trait Trait: srml_system::Trait {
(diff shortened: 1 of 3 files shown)


tokens/src/lib.rs
```diff
@@ -1,21 +1,45 @@
...
+// #3295 https://github.com/paritytech/substrate/issues/3295
+use srml_system::{self as system, ensure_signed};
+
 use traits::MultiCurrency;
 
-pub trait Trait: system::Trait {
-	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+pub trait Trait: srml_system::Trait {
+	type Event: From<Event<Self>> + Into<<Self as srml_system::Trait>::Event>;
+	type Balance: Parameter + Member + SimpleArithmetic + Codec + Default + Copy + MaybeSerializeDeserialize + Debug;
```
xlc:
```suggestion
	type Balance: Parameter + Member + SimpleArithmetic + Default + Copy + MaybeSerializeDeserialize;
```

`Parameter` already includes `Codec` and `Debug`.

tokens/src/lib.rs
```diff
@@ -1,21 +1,45 @@
...
 decl_storage! {
 	trait Store for Module<T: Trait> as Tokens {
+		/// The total issuance of a token type;
+		pub TotalIssuance get(fn total_issuance) build(|config: &GenesisConfig<T>| {
+			let issuance = config.initial_balance * (config.endowed_accounts.len() as u32).into();
+			config.tokens.iter().map(|id| (id.clone(), issuance)).collect::<Vec<_>>()
+		}): map T::CurrencyId => T::Balance;
 
+		/// The balance of a token type under an account.
+		pub Balance get(fn balance): double_map T::CurrencyId, twox_128(T::AccountId) => T::Balance;
```
xlc:
```suggestion
		pub Balance get(fn balance): double_map T::CurrencyId, blake2_256(T::AccountId) => T::Balance;
```

blake2_256 to be safe

tokens/src/lib.rs
```diff
@@ -25,9 +49,88 @@ decl_module! {
...
+		);
+
+		<TotalIssuance<T>>::mutate(currency_id, |v| *v -= amount);
+		<Balance<T>>::mutate(currency_id, who, |v| *v -= amount);
+
+		Ok(())
+	}
+
+	fn slash(currency_id: &Self::CurrencyId, who: &T::AccountId, amount: Self::Balance) -> Self::Balance {
+		let actual_amount = Self::balance(currency_id, who).max(amount);
```
xlc:
```suggestion
		let actual_amount = Self::balance(currency_id, who).min(amount);
```

traits/src/lib.rs
```diff
@@ -1,3 +1,45 @@
...
+	fn total_inssuance(currency_id: &Self::CurrencyId) -> Self::Balance;
+
+	/// The combined balance of `who` under `currency_id`.
+	fn balance(currency_id: &Self::CurrencyId, who: &AccountId) -> Self::Balance;
+
+	// Public mutables
+
+	/// Transfer some amount from one account to another.
+	fn transfer(
+		currency_id: &Self::CurrencyId,
```
xlc: `CurrencyId` already is `Copy` so you can just pass it by value

xlc:
For next phrase:
- use `decl_error` and avoid string errors
- make tokens module works with `Currency` / `Balances` module

shaopengw:
Will open other PRs for improvements.

As the ongoing "imbalances" types development would change most function signatures, unit tests would be added in other PRs along with these change.

Issue #4, opened 2019-11-01
Tokens Module


shaopengw:
## Goal

Create a `orml-tokens` runtime module to support multiple types of token.

### Version 1 - Basic tokens functionalities

-  [x] Minimal `trait` abstraction `MultiCurrency`.
  - `transfer`
  - `mint`
  - `burn`
  - `slash`
- [x] Runtime module.
  - Impl `MultiCurrency`.
  - Dispatchable `transfer` extrinsic
- [x] Unit test.

### Version 2 - Compatibility with `srml-balance` module

- [x] Be compatible with `srml-balance` and `Currency`, which would be used for native token.

~~### Optional - Better issuance control via "imbalance"~~

~~- [ ] "imbalance" types represents `balance` change of tokens.~~
~~- [ ] Add imbalance types and rebalance feature to `MultiCurrency` and `orml-tokens`.~~

xlc:
More things to be considered later:

- events
- error type
- weight / transaction fee
- existential deposits

xlc:
I am also thinking about adding a `BaseCurrency` trait that is similar to `MultiCurrency` except it is for single currency.

We don't need all the advanced features of `Currency` most of the time. `BaseCurrency` is usually enough. This allow new implementation of `BaseCurrency` easy.

PR #5, opened 2019-11-01
WIP - Imbalance

shaopengw:
Introduce "imbalance" for tokens module #4.

- [x] PoC of imbalance traits and types without `Subtrait`.
- [ ] Imbalance types `split`, `merge` functions etc.
- [x] Refactor `MultiCurrency`'s `mint/burn/slash`.
- [ ] Imbalance unit tests.
- [ ] Tokens module unit tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,24 +1,7 @@
-[package]
-name = "orml"
-version = "0.0.1"
-authors = ["Laminar Developers <hello@laminar.one>"]
-edition = "2018"
-
-[[bin]]
-name = "orml"
-path = "src/main.rs"
-
-[dependencies]
-oracle = { package = "orml-oracle", path = "oracle", default-features = false }
-tokens = { package = "orml-tokens", path = "tokens", default-features = false }
-traits = { package = "orml-traits", path = "traits", default-features = false }
-utilities = { package = "orml-utilities", path = "utilities", default-features = false }
-
-[features]
-default = ["std"]
-std = [
-	"oracle/std",
-	"tokens/std",
-	"traits/std",
-	"utilities/std",
+[workspace]
+members = [
+	"oracle",
+	"tokens",
+	"traits",
+	"utilities",
 ]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -1,24 +1,7 @@
...
-utilities = { package = "orml-utilities", path = "utilities", default-features = false }
-
-[features]
-default = ["std"]
-std = [
-	"oracle/std",
-	"tokens/std",
-	"traits/std",
-	"utilities/std",
+[workspace]
```
xlc: I had some issue with `cargo check --all  --no-default-features` with this setup. Can you make sure the targets in Makefile works? Or if you can a better way to do check and run tests.

tokens/src/imbalances.rs
```diff
@@ -19,44 +19,68 @@ impl<T: Trait> Rebalance<T::CurrencyId, T::Balance> for RebalanceNegative<T> {
 	}
 }
 
+#[derive(Debug, PartialEq, Eq)]
```
xlc: use `RuntimeDebug` for derive https://github.com/paritytech/substrate/pull/3821

tokens/src/imbalances.rs
```diff
@@ -19,44 +19,68 @@ impl<T: Trait> Rebalance<T::CurrencyId, T::Balance> for RebalanceNegative<T> {
...
 	}
 }
 
+impl<T: Trait> Drop for PositiveImbalance<T> {
+	fn drop(&mut self) {
+		<Self as Imbalance>::Rebalance::rebalance(self.currency_id, self.amount);
+	}
+}
+
+#[derive(Debug, PartialEq, Eq)]
```
xlc: RuntimeDebug

Cargo.toml
shaopengw: @xlc Will resolve this err. I'm thinking of adding a example crate for both `orml` usage demo and integration check/tests purpose, we could probably check `--no-default-features` in the example crate.

shaopengw:
As discussed offline with @xlc , introducing imbalance types would added API/maintenance complexities bigger that their benefits(at least at this stage). Will park this PR for now until we found it really necessary in the future.

PR #6, opened 2019-11-05
Tokens module unit tests

shaopengw:
* Unit tests for tokens module #4 .
* Example module.
* Some minor fixes.


Makefile
```diff
@@ -1,5 +1,5 @@
-check: githooks
-	cargo check --no-default-features
+check-no-std: githooks
+	cd example && cargo check --no-default-features && cd ..
```
xlc:
```suggestion
	cd example && cargo check --no-default-features
```

make spawn new shell so no need cd back

Makefile
```diff
@@ -1,5 +1,5 @@
-check: githooks
-	cargo check --no-default-features
+check-no-std: githooks
```
xlc: I still want to just keep it as `make check` to save typings

tokens/src/mock.rs
```diff
@@ -0,0 +1,118 @@
...
+
+	#[allow(dead_code)]
+	pub fn balances(mut self, account_ids: Vec<AccountId>, initial_balance: Balance) -> Self {
+		self.endowed_accounts = account_ids;
+		self.initial_balance = initial_balance;
+		self
+	}
+
+	pub fn one_hundred_for_alice_n_bob(mut self) -> Self {
+		self.endowed_accounts = vec![ALICE, BOB];
```
xlc: this can call `self.balances` to make it used and remove `#[allow(dead_code)]`

tokens/src/mock.rs
```diff
@@ -0,0 +1,118 @@
...
+		Self {
+			currency_id: TEST_TOKEN_ID,
+			endowed_accounts: vec![0],
+			initial_balance: 0,
+		}
+	}
+}
+
+impl ExtBuilder {
+	#[allow(dead_code)]
```
xlc: remove it if not used

tokens/src/mock.rs
```diff
@@ -0,0 +1,110 @@
...
```
xlc:
```suggestion
```

PR #13, opened 2019-11-07
Tokens module: impl `MultiCurrencyExtended`

shaopengw:


traits/Cargo.toml
```diff
@@ -12,6 +12,7 @@ support = { package = "srml-support", git = "https://github.com/paritytech/subst
 system = { package = "srml-system", git = "https://github.com/paritytech/substrate.git", default-features = false }
 runtime_io = { package = "sr-io", git = "https://github.com/paritytech/substrate.git", default-features = false }
 rstd = { package = "sr-std", git = "https://github.com/paritytech/substrate.git", default-features = false }
+num-traits = { version = "0.2.8", default-features = false }
```
xlc: add `num-traits/std` to std features

tokens/src/lib.rs
```diff
@@ -10,14 +13,23 @@ use srml_support::{decl_error, decl_event, decl_module, decl_storage, ensure, Pa
...
-use traits::MultiCurrency;
+use traits::{arithmetic, arithmetic::Signed, MultiCurrency, MultiCurrencyExtended};
 
 mod mock;
 mod tests;
 
 pub trait Trait: srml_system::Trait {
 	type Event: From<Event<Self>> + Into<<Self as srml_system::Trait>::Event>;
 	type Balance: Parameter + Member + SimpleArithmetic + Default + Copy + MaybeSerializeDeserialize;
+	type Amount: arithmetic::Signed
```
xlc: you `use arithmetic::Signed` and still `arithmetic::Signed` here?

shaopengw: Aha, forgot to change it after adding `Signed` to support `abs()` and `is_positive()`.

Issue #8, opened 2019-11-05
Overall Design


xlc:
- utilities
  - Types
    - [x] struct LinkedList
      - https://github.com/xlc/substrate-kitties/blob/master/runtime/src/linked_item.rs
- traits
  - Types
    - [x] trait BasicCurrency
      - type Balance
      - fn transfer
      - fn deposit
      - fn withdraw
    - [x] trait BasicCurrencyExtended: MultiCurrency
      - type Amount: TryInto\<Balance> + TryFrom\<Balance>
      - fn update_balance(who: AccountId, amount: Amount)
    - [x] trait MultiCurrency (see #4)
      - type CurrencyId
      - type Balance
      - fn transfer
      - fn deposit
      - fn withdraw
    - [x] trait MultiCurrencyExtended: MultiCurrency
      - type Amount: TryInto\<Balance>, TryFrom\<Balance>
        - Signed version of `Balance`
      - fn update_balance(who: AccountId, currency_id: CurrencyId, amount: Amount)
    - [x] trait OnNewData\<Key, Value>
      - fn on_new_data(key: Key, value: Value)
    - [x] trait DataProvider\<Key, Value>
      - fn get(key: Key) -> Option\<Value>
    - [x] trait PriceProvider<CurrencyId, Price>
      - fn get_price(base: CurrencyId, quote: CurrencyId): Option\<Price>
    - [x] struct AuctionInfo
      - bid: Option<(AccountId, Balance)>
      - end: Option\<BlockNumber>
    - [x] trait Auction
      - type AuctionId
      - type Balance
      - fn auction_info(id: AuctionId): AuctionInfo
      - fn update_auction(id: AuctionId, info: AuctionInfo)
      - fn new_auction(start: BlockNumber, end: Option\<BlockNumber>): AuctionId
    - [x] struct OnNewBidResult
      - accept_bid: bool
      - auction_end: Optio<Option\<BlockNumber>>
    - [x] trait AuctionHandler
      - fn on_new_bid(now: BlockNumber, id: AuctionId, new_bid: (AccountId, Balance), last_bid: Option<(AccountId, Balance)>): OnNewBidResult
        - called when new bid is received
        - use return value to determine if the bid should be accepted and update auction end time
        - this should reserved current winner money and refund previous winner
      - fn on_auction_ended(id: AuctionId, winner: Option<(AccountId, Balance)>)
- [x] tokens
  - Trait
    - CurrencyId
  - Call
[...]

PR #22, opened 2019-11-11
Currencies Module

### LINKED ISSUE #4: Tokens Module

shaopengw:
impl currencies module in #8, closes #4


currencies/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+		Transferred(CurrencyId, AccountId, AccountId, Balance),
+	}
+);
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		fn deposit_event() = default;
+
+		/// Transfer some balance to another account.
+		pub fn transfer(
```
xlc: should have transfer handle both native and non native currency and use `T:: NativeCurrency ` for native transfer

currencies/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+type BalanceOf<T> = <<T as Trait>::MultiCurrency as MultiCurrency<<T as srml_system::Trait>::AccountId>>::Balance;
+type CurrencyIdOf<T> = <<T as Trait>::MultiCurrency as MultiCurrency<<T as srml_system::Trait>::AccountId>>::CurrencyId;
+type ErrorOf<T> = <<T as Trait>::MultiCurrency as MultiCurrency<<T as srml_system::Trait>::AccountId>>::Error;
+
+type AmountOf<T> = <<T as Trait>::MultiCurrency as MultiCurrencyExtended<<T as srml_system::Trait>::AccountId>>::Amount;
+
+pub trait Trait: system::Trait {
+	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
+	type MultiCurrency: MultiCurrencyExtended<Self::AccountId>;
+	type GetNativeCurrencyId: Get<CurrencyIdOf<Self>>;
```
xlc: `type NativeCurrency: BasicCurrency`

currencies/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+impl<T: Trait> MultiCurrency<T::AccountId> for Module<T> {
+	type Balance = BalanceOf<T>;
+	type CurrencyId = CurrencyIdOf<T>;
+	type Error = ErrorOf<T>;
+
+	fn total_issuance(currency_id: Self::CurrencyId) -> Self::Balance {
+		T::MultiCurrency::total_issuance(currency_id)
```
xlc: all those methods should check if currency_id == native and delegate to `T::NativeCurrency` instead

xlc:
Please merge master for CI fix.

shaopengw:
The `BasicCurrency` impl for `srml_support::traits::Currency` has to be in `traits` module because of the orphan rule.

With `impl<AccountId, T: SrmlCurrency<AccountId>> BasicCurrency<AccountId> for T { ... }`, implementing `BasicCurrency` for a generic `T: srml_support::traits::Currency` would conflict with the `BasicCurrency` impl for [`Currency` struct](https://github.com/laminar-protocol/open-runtime-module-library/blob/0e13df4278e537b52c37f7fdd0a83241b25cf51b/currencies/src/lib.rs#L172), because of the restriction [E0119](https://doc.rust-lang.org/error-index.html#E0119).

A possible solution is adding another trait bound for the generic `T`, as long as `Currency` struct doesn't impl it, there would be no conflicts. `dispatch::ModuleErrorMetadata` looks like good choice, since runtime modules implement it and there is no additional generic type.

xlc:
I am actually thinking about something like this

```
struct BasicCurrencyAdpater<C>;

impl<AccountId, C> BasicCurrency for BasicCurrencyAdpater<C>
where C: Currency<AccountId>
```

and to use

```
impl auction::Triat for Runtime {
    type Currency = BasicCurrencyAdpater<Balances>;
}
```

shaopengw:
@xlc Updated to adapter implementation.

currencies/src/lib.rs
```diff
@@ -218,4 +223,51 @@ where
...
+	fn balance(who: &AccountId) -> Self::Balance {
+		T::total_balance(who)
+	}
+
+	fn transfer(from: &AccountId, to: &AccountId, amount: Self::Balance) -> result::Result<(), Self::Error> {
+		T::transfer(from, to, amount, ExistenceRequirement::AllowDeath)
+	}
+
+	fn deposit(who: &AccountId, amount: Self::Balance) -> result::Result<(), Self::Error> {
+		let imbalance = T::deposit_creating(who, amount);
```
xlc: `let _ = T::deposit_creating(who, amount);` should be the same

currencies/src/lib.rs
```diff
@@ -218,4 +223,51 @@ where
...
+	}
+
+	fn withdraw(who: &AccountId, amount: Self::Balance) -> result::Result<(), Self::Error> {
+		T::withdraw(
+			who,
+			amount,
+			WithdrawReason::Transfer.into(),
+			ExistenceRequirement::AllowDeath,
+		)
+		.map(|imbalance| {
```
xlc: I don't think you have to call drop, so just `.map(|_| ())` should be good enough

currencies/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+			who,
+			amount,
+			WithdrawReason::Transfer.into(),
+			ExistenceRequirement::AllowDeath,
+		)
+		.map(|_| ())
+	}
+
+	fn slash(who: &AccountId, amount: Self::Balance) -> Self::Balance {
+		let (imbalance, gap) = T::slash(who, amount);
```
xlc:
```suggestion
		let (_, gap) = T::slash(who, amount);
```

PR #27, opened 2019-11-14
LinkedList enumerator

### LINKED ISSUE #21: LinkedItem enumerator
- [ ] Add `enumerate(key: &Key)` to be able to iterate all items for a given key. Similar to [`LinkedStorageMap::enumerate`](https://github.com/paritytech/substrate/blob/a86bb37b7c9eb7d583e5554fc0790050beaa14d5/srml/support/src/storage/generator/linked_map.rs#L338)
- [ ] Add `take_all(key: &Key)` to return an enumerator that iterates all items for a given key and remove all items at same time. This enumerator should implement `Drop` to delete all un-iterated  items.

shaopengw:
Closes #21

### DIFF
--- a/utilities/src/linked_item.rs
+++ b/utilities/src/linked_item.rs
@@ -9,6 +9,12 @@ pub struct LinkedItem<Item> {
 	pub next: Option<Item>,
 }
 
+impl<Item> Default for LinkedItem<Item> {
+	fn default() -> Self {
+		LinkedItem { prev: None, next: None }
+	}
+}
+
 pub struct LinkedList<Storage, Key, Item>(rstd::marker::PhantomData<(Storage, Key, Item)>);
 
 impl<'a, Storage, Key, Value> LinkedList<Storage, Key, Value>
@@ -26,7 +32,7 @@ where
 	}
 
 	fn read(key: &Key, value: Option<Value>) -> LinkedItem<Value> {
-		Storage::get(&(key.clone(), value)).unwrap_or_else(|| LinkedItem { prev: None, next: None })
+		Storage::get(&(key.clone(), value)).unwrap_or_else(|| Default::default())
 	}
 
 	fn write(key: &Key, value: Option<Value>, item: LinkedItem<Value>) {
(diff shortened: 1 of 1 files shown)


utilities/src/linked_item.rs
```diff
@@ -74,8 +87,68 @@ where
...
+// 	Storage: StorageMap<(Key, Option<Value>), LinkedItem<Value>, Query = Option<LinkedItem<Value>>>,
+// {
+// 	fn drop(&mut self) {
+// 		if !self.should_take {
+// 			return
+// 		}
+
+// 		while self.next() != None {}
+// 	}
+// }
```
shaopengw: This straightforward implementation doesn't work:
> error[E0366]: Implementations of Drop cannot be specialized

shaopengw:
`Drop` could not be implemented in this straightforward way.

> error[E0366]: Implementations of Drop cannot be specialized

https://github.com/laminar-protocol/open-runtime-module-library/blob/3ddd63ec16dade262eed7056934c7ad1a9ec8f64/utilities/src/linked_item.rs#L136-L150

utilities/src/linked_item.rs
```diff
@@ -8,9 +9,15 @@ pub struct LinkedItem<Item> {
 	pub next: Option<Item>,
 }
 
+impl<Item> Default for LinkedItem<Item> {
```
xlc: can you derive Default?

utilities/src/linked_item.rs
```diff
@@ -20,12 +27,18 @@ where
...
 
-	fn write_head(account: &Key, item: LinkedItem<Value>) {
-		Self::write(account, None, item);
+	fn write_head(key: &Key, item: LinkedItem<Value>) {
+		Self::write(key, None, item);
 	}
 
 	fn read(key: &Key, value: Option<Value>) -> LinkedItem<Value> {
-		Storage::get(&(key.clone(), value)).unwrap_or_else(|| LinkedItem { prev: None, next: None })
+		Storage::get(&(key.clone(), value)).unwrap_or_else(|| Default::default())
```
xlc:
```suggestion
		Storage::get((&key, value)).unwrap_or_else(Default::default)
```

utilities/src/linked_item.rs
shaopengw: rustc requires `Value: core::default::Default` bound to use `Default::default` in LinkedList impl, might makes `Value` kind of restricted?

xlc: ah ok

PR #66, opened 2019-12-23
allow specify balances for each users and each tokens

ermalkaleci:


tokens/src/lib.rs
```diff
@@ -31,30 +34,42 @@ pub trait Trait: frame_system::Trait {
...
 		pub TotalIssuance get(fn total_issuance) build(|config: &GenesisConfig<T>| {
-			let issuance = config.initial_balance * (config.endowed_accounts.len() as u32).into();
-			config.tokens.iter().map(|id| (id.clone(), issuance)).collect::<Vec<_>>()
+			config
+				.endowed_accounts
+				.iter()
+				.map(|(_, currency_id, initial_balance)| (currency_id, initial_balance))
+				.fold(BTreeMap::<T::CurrencyId, T::Balance>::new(), |mut acc, (currency_id, initial_balance)| {
+					if let Some(issuance) = acc.get_mut(currency_id) {
+						if let Some(sum) = issuance.checked_add(initial_balance) {
```
xlc: Too bad Rust does not have `panic_add`, but `checked_add` with `expect` should achieve the same thing.
`*issuance = issuance.checked_add(initial_balance).expect("total issuance cannot overflow when building genesis");`

PR #67, opened 2019-12-25
update ended auctions cleanup

ermalkaleci:


auction/src/lib.rs
```diff
@@ -117,7 +102,18 @@ decl_error! {
...
 }
 
-impl<T: Trait> Module<T> {}
+impl<T: Trait> Module<T> {
+	fn _on_finalize(now: T::BlockNumber) {
+		<AuctionEndTimeList<T>>::take_all(&now).for_each(|auction_id| {
+			if let Some(auction) = <Auctions<T>>::take(auction_id) {
+				T::Handler::on_auction_ended(auction_id, auction.bid.clone());
+				<AuctionsCount<T>>::mutate(|n| {
+					*n = n.checked_sub(&T::AuctionId::one()).unwrap_or(T::AuctionId::zero())
```
xlc: use `saturating_sub`

I am thinking if we need to keep this updated one by one. the fast way is keep a count and sub the count outside of the loop.
But it will mean if the `on_auction_ended` handler trying to read `AuctionsCount` it will read the old value, which may be fine.
In anycase, the mutate should be before `on_auction_ended`, otherwise it will be slow and inaccurate.

PR #83, opened 2020-01-31
Oracle RPC

### LINKED ISSUE #79: Add a RPC API to query combined oracle value
As the oracle value is lazily combined from different source, querying chain state `oracle/values` cannot always get the current combined data.

The oracle module could provide a RPC API helper, to provide fresh combined data, so that front end don't have to re-implement the combining logic.

shaopengw:
It works, finally. More TODO:
- [x] unit tests

Closes #79

### DIFF
--- a/oracle/rpc/src/lib.rs
+++ b/oracle/rpc/src/lib.rs
@@ -1,0 +1,72 @@
+use std::sync::Arc;
+use codec::Codec;
+use sp_blockchain::HeaderBackend;
+use jsonrpc_core::{Error as RpcError, ErrorCode, Result};
+use jsonrpc_derive::rpc;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use sp_api::ProvideRuntimeApi;
+pub use orml_oracle_rpc_runtime_api::OracleApi as OracleRuntimeApi;
+pub use self::gen_client::Client as OracleClient;
+
+#[rpc]
+pub trait OracleApi<BlockHash, Key, Value> {
+	#[rpc(name = "oracle_getNoOp")]
+	fn get_no_op(
+		&self,
+		key: Key,
+		at: Option<BlockHash>
+	) -> Result<Option<Value>>;
+}
+
+/// A struct that implements the [`OracleApi`].
+pub struct Oracle<C, B> {
+	client: Arc<C>,
+	_marker: std::marker::PhantomData<B>,
+}
+
+impl<C, B> Oracle<C, B> {
+	/// Create new `Oracle` with the given reference to the client.
+	pub fn new(client: Arc<C>) -> Self {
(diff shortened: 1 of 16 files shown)


utilities/src/fixed128.rs
```diff
@@ -7,12 +7,12 @@ use sp_runtime::{
...
 
 #[cfg(feature = "std")]
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 
 /// An unsigned fixed point number. Can hold any value in the range [0, 340_282_366_920_938_463_464]
 /// with fixed point accuracy of 10 ** 18.
 #[derive(Encode, Decode, Default, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "std", derive(Deserialize))]
```
xlc: Serialize and Deserialize won't match like this and it could cause issue on chain spec generation

oracle/rpc/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+use jsonrpc_derive::rpc;
+pub use orml_oracle_rpc_runtime_api::OracleApi as OracleRuntimeApi;
+use sp_api::ProvideRuntimeApi;
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use std::sync::Arc;
+
+#[rpc]
+pub trait OracleApi<BlockHash, Key, Value> {
+	#[rpc(name = "oracle_getNoOp")]
```
xlc:
```suggestion
	#[rpc(name = "orml_oracle_getValue")]
```

why call it getNoOp? RPC cannot have side effects on on-chain storage. I think getValue is fine.

utilities/src/fixed128.rs
shaopengw: Added `Deserialize` impl.

oracle/rpc/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+pub use orml_oracle_rpc_runtime_api::OracleApi as OracleRuntimeApi;
+use sp_api::ProvideRuntimeApi;
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use std::sync::Arc;
+
+#[rpc]
+pub trait OracleApi<BlockHash, Key, Value> {
+	#[rpc(name = "oracle_getValue")]
+	fn get_no_op(&self, key: Key, at: Option<BlockHash>) -> Result<Option<Value>>;
```
xlc: rename this as well

PR #109, opened 2020-03-05
Vesting module documentation

shaopengw:


shaopengw:
The failing check is caused by incompatibility between latest `quote@1.0.3` and `failure-derive`, not related to this PR. Users of ORML won't be affected as they have `Cargo.lock`.

PR #120, opened 2020-03-11
update T::AccountId to use blake2_128_concat

ermalkaleci:


xlc:
Oracle RawValues required account with private key to write so `twox_64_concat` is safe.
And the other one is just test.

PR #121, opened 2020-03-12
Add is_positive and is_negative fn to Fixed128.

shaopengw:


shaopengw:
> should just implement `num_traits::sign::Signed`

`num_traits::sign::Signed` must have `abs` fn, but `Fixed128` could only have `saturating_abs`.

PR #127, opened 2020-03-20
replace div to mul by signum

ermalkaleci:


xlc:
A unit test catch this issue?

ermalkaleci:
@xlc no. I discovered it while writing tests for offchain worker. It will panic if signum becomes zero

xlc:
can you add one?

PR #153, opened 2020-05-08
update trait for benchmarking

wangjj9219:
1. <s>It's unable to register type directly in benchmarking context, and acala need to convert CurrencyId in tokens.Trait to OracleKey(CurrencyId) in oracle.Trait, so add trait `TryInto<u16>` to tokens.Trait::CurrencyId </s>
2. impl new trait `DataProviderExtended` for oracle

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,9 +30,12 @@ jobs:
         toolchain: nightly-2020-02-18
         components: rustfmt
         target: wasm32-unknown-unknown
+        override: true
         default: true
     - name: Install Wasm toolchain
       run: rustup target add wasm32-unknown-unknown
+    - name: Install cargo-unleash
+      run: cargo install cargo-unleash --version 1.0.0-alpha.10 #FIXME: change to release once it's ready
     - name: Check format
       run: make dev-format-check
     - name: Update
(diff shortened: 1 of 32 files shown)


shawntabrizi:
If there is some feature missing in the benchmarking pipeline please do feel free to open an issue.

We also ran into some trait bound problems, but I have found that is mostly a problem with our own pallets not exposing APIs that would make them truly autonomous in a generic runtime.

shaopengw:
I'm doing benchmarks on laminar-chain at the moment, and also run into quite a bit trait types problems. Particularly Balance/Currency/Price types etc, adding more trait bounds just for benchmarking reason might a bit overkill, instead you could add helper methods under `runtime-benchmarks` features, like [here](https://github.com/laminar-protocol/laminar-chain/blob/cf73272bdd48f04490f7d5252a8f7f4cf326a0f5/modules/margin-protocol/src/benchmarking_utilities.rs#L1-L11).

wangjj9219:
> I'm doing benchmarks on laminar-chain at the moment, and also run into quite a bit trait types problems. Particularly Balance/Currency/Price types etc, adding more trait bounds just for benchmarking reason might a bit overkill, instead you could add helper methods under `runtime-benchmarks` features, like [here](https://github.com/laminar-protocol/laminar-chain/blob/cf73272bdd48f04490f7d5252a8f7f4cf326a0f5/modules/margin-protocol/src/benchmarking_utilities.rs#L1-L11).

you are right. And I introduce a new trait `DataProviderExtended` to expose feed_value so that avoid extra changes

PR #154, opened 2020-05-08
Alpha 6 [NOT FOR MERGING]

xlc:


JoshOrndorff:
We tried this on the node template too and were ultimately unsuccessful. Cargo will pull those alpha.7 packages despite the `=` sign if you add a new crate that isn't in `Cargo.lock`.

xlc:
@JoshOrndorff thanks for the info..... seems like we really need semvar in Substrate...

PR #162, opened 2020-05-14
Use Clippy in CI

### LINKED ISSUE #19: Add cargo clippy to CI pipeline
### Goal:

Add `cargo clippy` to Github Action pipeline to perform extra linting. 

### Background:

[cargo clippy](https://github.com/rust-lang/rust-clippy) is a Rust linter that designed to catch common mistakes and we would like to run it in our PR pipeline.

### Actions:

Run and configure `cargo clippy`.

Document all the issues found and fix all the easy one.

Update Github Actions PR pipeline to run `cargo clippy` to prevent regression.

KiChjang:
Fixes #19.

We may also want to configure clippy so that it errors on every warning in CI.

### DIFF
--- a/vesting/src/lib.rs
+++ b/vesting/src/lib.rs
@@ -21,7 +21,6 @@
 //! - `update_vesting_schedules` - Update all vesting schedules under an account, `root` origin required.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-
 // Disable the following two lints since they originate from an external macro (namely decl_storage)
 #![allow(clippy::redundant_closure_call, clippy::string_lit_as_bytes)]
 
--- a/vesting/src/lib.rs
+++ b/vesting/src/lib.rs
@@ -22,6 +22,9 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+// Disable the following two lints since they originate from an external macro (namely decl_storage)
+#![allow(clippy::redundant_closure_call, clippy::string_lit_as_bytes)]
+
 use codec::{Decode, Encode, HasCompact};
 use frame_support::{
 	decl_error, decl_event, decl_module, decl_storage, ensure,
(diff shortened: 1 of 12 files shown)


vesting/src/lib.rs
```diff
@@ -22,6 +22,9 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+// Disable the following two lints since they originate from an external macro (namely decl_storage)
+#![allow(clippy::redundant_closure_call, clippy::string_lit_as_bytes)]
+
```
JoshOrndorff: @thiolliere Is this something we can fix in Substrate?  Or is it best for people building on Substrate to suppress the warning like this?

JoshOrndorff: It is used to suppress warnings similar to this
```
warning: calling `as_bytes()` on a string literal
  --> pallets/adding-machine/src/lib.rs:17:1
   |
17 | / decl_storage! {
18 | |     trait Store for Module<T: Trait> as AddingMachine {
19 | |         Sum get(fn sum): u32;
20 | |     }
21 | | }
   | |_^
   |
   = note: `#[warn(clippy::string_lit_as_bytes)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#string_lit_as_bytes
   = note: this warning originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

xlc: I think this is the cause https://github.com/paritytech/substrate/blob/ea501c2d61d27e7f0f76f93e5fa7e5c90882da2f/frame/support/procedural/src/storage/storage_struct.rs#L109

thiolliere: yes we have fixed some in the past I'll take a look to fix those

PR #187, opened 2020-05-27
update-fixedu128

qwer951123:
* break from_ration to saturating_from_ration and checked_from_ration.
* break from_natural to saturating_from_integer and checked_from_integer.
* rename from_parts to from_inner.
* implement Zero, One, Add, Div, Mul, Sub traits.
* add saturating_mul_int, saturating_div_int.
* add turns, frac, cell, floor, round.


wangjj9219:
@shaopengw the interface of FixedU128 has updated, check it if Laminar updates orml

PR #190, opened 2020-05-28
Add dispatch origin, fix remove update logic

xlc:


gradually-update/src/lib.rs
```diff
@@ -117,41 +121,47 @@ impl<T: Trait> Module<T> {
...
-		}
+				Self::deposit_event(RawEvent::GraduallyUpdateBlockNumber(now, update.key.clone(), value));
+
+				keep
+			})
+			.collect::<Vec<_>>();
 
 		// gradually_update has finished. Remove it from GraduallyUpdates.
-		if gradually_updates.len() < GraduallyUpdates::decode_len().unwrap_or_default() {
+		if gradually_updates.len() < initial_count {
```
zjb0807, line 164: `gradually_updates.len() < initial_count ` is always false.
`gradually_updates` need filter.

xlc: `gradually_updates` is reassigned with the filtered value so it should be good

PR #200, opened 2020-06-03
add authority module

wangjj9219:


authority/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+pub struct EnsureDelayed<Delay, Inner, BlockNumber>(sp_std::marker::PhantomData<(Delay, Inner, BlockNumber)>);
+
+impl<
+		O: Into<Result<DelayedOrigin<BlockNumber, O>, O>> + From<DelayedOrigin<BlockNumber, O>>,
+		Delay: Get<BlockNumber>,
+		Inner: EnsureOrigin<O>,
+		BlockNumber: PartialOrd,
+	> EnsureOrigin<O> for EnsureDelayed<Delay, Inner, BlockNumber>
+{
+	type Success = ();
```
xlc, line 42: Something like this to keep Success type (not sure syntax) 

`type Success = Inner::Success;`

authority/src/lib.rs
```diff
@@ -28,7 +30,8 @@ pub struct DelayedOrigin<BlockNumber, Origin> {
 }
 
 /// Origin for the authority module.
-pub type Origin<T> = DelayedOrigin<<T as system::Trait>::BlockNumber, <T as frame_system::Trait>::Origin>;
+pub type Origin<T> =
+	DelayedOrigin<<T as system::Trait>::BlockNumber, system::RawOrigin<<T as frame_system::Trait>::AccountId>>;
```
xlc, line 36:
```suggestion
	DelayedOrigin<<T as system::Trait>::BlockNumber, system::RawOrigin<<T as system::Trait>::AccountId>>;
```

PR #202, opened 2020-06-04
Upgrade to rc2.

dvc94ch:
This was simpler than expected. Ran `cargo upgrade --all` and `cargo test`.


xlc:
Thanks for the PR. rc1 -> rc2 is have a relatively small changeset. The next upgrade will be harder...

PR #215, opened 2020-06-26
Change Option<bool> storage to Option<()>.

### LINKED ISSUE #210: Replace `Option<bool>` storage to `Option<()>`
Depends on https://github.com/paritytech/substrate/pull/6364

Substrate now supports empty storage value so we can change `Option<bool>` storage to `Option<()>`.

shaopengw:
Close #210


JoshOrndorff:
Have you tried reading state from offchain with the `()` storage value? Like from polkadot-js API for example? I have not yet.

xlc:
polkadot.js should support it, otherwise I don't see any major blockers for us to fix it.

Issue #245, opened 2020-08-07
MultiDataProvider


xlc:
Make orml-oracle instantiable 

- Types
  - `trait MultiDataProvider<ProviderId, Key, Value>`
    - `fn get(source: ProviderId, key: &Key) -> Option<Value>`
  - `macro! create_midean_value_data_provider(TypeName:ident, $(Providers:ty),+)`
    - takes list of type that `impl DataProvider`
    - create a type `struct $TypeName<MinimumCount: Get<u32>, ExpiresIn: Get<Moment>, Key, Value: Ord>`
      -  `impl DataProvider` like the `DefaultCombineData` but read values from `$Providers`
- Runtime
  - `AcalaDataProvider: orml_oracle::Instance1`
  - `BandDataProvider: orml_oracle::Instance2`
  - `enum DataProviderId`
    - `AcalaDataProvier`
    - `BandDataProvider`
    - `AggergatedDataProvider`
  - `create_midean_value_data_provider(AggregatedDataProvider, AcalaDataProvider, BandDataProvider) `
  - `struct MultiDataProviderImpl`
    - `impl MultiDataProvider<DataProviderId, CurrencyId, Price>`
  - `Prices::Source = AggregatedDataProvider<Two, OneHour, CurrencyId, Price>`

shaopengw:
Closed by #256

PR #256, opened 2020-08-20
Make orml oracle instantiable

prin-r:
[see specification](https://github.com/open-web3-stack/open-runtime-module-library/issues/245)

Orml

- ✅ Make orml-oracle instantiable
- ✅ trait MultiDataProvider<ProviderId, Key, Value>
  - fn get(source: ProviderId, key: &Key) -> Option<Value>
- ✅ macro! create_midean_value_data_provider(TypeName:ident, $(Providers:ty),+)
  - takes list of type that impl DataProvider
  - create a type struct $TypeName<MinimumCount: Get<u32>, ExpiresIn: Get<Moment>, Key, Value: Ord>
  - impl DataProvider like the DefaultCombineData but read values from $Providers

### DIFF
--- a/traits/src/data_provider.rs
+++ b/traits/src/data_provider.rs
@@ -1,4 +1,5 @@
 use sp_runtime::DispatchResult;
+use sp_std::vec::Vec;
 
 /// Data provider with ability to provide data with no-op, and provide all data.
 pub trait DataFeeder<Key, Value, AccountId>: DataProvider<Key, Value> {
--- a/traits/src/data_provider.rs
+++ b/traits/src/data_provider.rs
@@ -1,0 +1,106 @@
+use sp_runtime::DispatchResult;
+
+/// Data provider with ability to provide data with no-op, and provide all data.
+pub trait DataFeeder<Key, Value, AccountId>: DataProvider<Key, Value> {
+	/// Provide a new value for a given key from an operator
+	fn feed_value(who: AccountId, key: Key, value: Value) -> DispatchResult;
+}
+
+/// A simple trait to provide data
+pub trait DataProvider<Key, Value> {
+	/// Get data by key
+	fn get(key: &Key) -> Option<Value>;
+}
+
+/// A simple trait to provide data from a given ProviderId
+pub trait MultiDataProvider<ProviderId, Key, Value> {
+	/// Provide a new value for given key and ProviderId from an operator
+	fn get(source: ProviderId, key: &Key) -> Option<Value>;
+}
+
+#[allow(dead_code)] // rust cannot defect usage in macro_rules
+fn median<T: Ord + Clone>(mut items: Vec<T>) -> Option<T> {
+	if items.is_empty() {
+		return None;
+	}
+
+	let mid_index = items.len() / 2;
+	// Won't panic as guarded items not empty case.
+	let (_, median, _) = items.partition_at_index(mid_index);
+	Some(median.clone())
+}
+
+#[macro_export]
+macro_rules! create_median_value_data_provider {
+	($name:ident, $key:ty, $value:ty, [$( $provider:ty ),*]) => {
+		pub struct $name;
+		impl DataProvider<$key, $value> for $name {
+			fn get(key: &$key) -> Option<$value> {
+				let mut values = vec![];
+				$(
+					if let Some(v) = <$provider>::get(&key) {
+						values.push(v);
+					}
+				)*
+				median(values)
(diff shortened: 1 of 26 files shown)


traits/src/lib.rs
```diff
@@ -56,3 +64,127 @@ pub enum Change<Value> {
...
+							timestamp: (x.timestamp + y.timestamp) / 2u64,
+						}),
+						_ => None,
+					}
+				} else {
+					Some(values[mid])
+				}
+			}
+
+			fn get_all_values() -> Vec<(CurrencyId, Option<TimestampedValue<Price, Moment>>)> {
```
shaopengw, line 140: Not 100% sure, should we also find median in `get_all_values`? @xlc

shaopengw:
Also needs unit tests to cover changes in this PR.

traits/src/lib.rs
```diff
@@ -56,3 +64,127 @@ pub enum Change<Value> {
...
+				)*
+
+				values.retain(|&x| x != None);
+				values.sort_by(|a, b| a.cmp(&b));
+				if values.len() == 0 {
+					return None;
+				}
+				let mid = values.len() / 2;
+				if values.len() % 2 == 0 {
+					match (values[mid - 1], values[mid]) {
```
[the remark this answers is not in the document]
prin-r: @shaopengw , I have added a comment as your suggested.
Now, the rest is to add units test.

prin-r:
@xlc , @shaopengw 

As discussed on discord, I'm handing this off to you guys. Please take this forward by adding tests. Let me know if I need to do anything.

traits/src/data_provider.rs
```diff
@@ -0,0 +1,138 @@
...
+/// Extended data provider to provide timestamped data by key with no-op, and
+/// all data.
+pub trait DataProviderExtended<Key, TimestampedValue> {
+	/// Get timestamped value by key
+	fn get_no_op(key: &Key) -> Option<TimestampedValue>;
+	/// Provide a list of tuples of key and timestamped value
+	fn get_all_values() -> Vec<(Key, Option<TimestampedValue>)>;
+}
+
+#[allow(dead_code)] // rust cannot defect usage in macro_rules
```
xlc, line 25:
```suggestion
#[allow(dead_code)] // rust cannot detect usage in macro_rules
```

traits/src/data_provider.rs
```diff
@@ -0,0 +1,138 @@
...
+	Some(median.clone())
+}
+
+#[macro_export]
+macro_rules! create_median_value_data_provider {
+	($name:ident, $key:ty, $value:ty, $timestamped_value:ty, [$( $provider:ty ),*]) => {
+		pub struct $name;
+		impl $crate::DataProvider<$key, $value> for $name {
+			fn get(key: &$key) -> Option<$value> {
+				let mut values = vec![];
```
xlc, line 43:
```suggestion
				let values = vec![ $(<$provider as $crate::DataProvider<$key, $value>>::get(&key)),* ];
```

traits/src/data_provider.rs
```diff
@@ -0,0 +1,138 @@
...
+					}
+				)*
+				$crate::data_provider::median(values)
+			}
+			fn get_all_values() -> Vec<($key, Option<$timestamped_value>)> {
+				let mut keys = sp_std::collections::btree_set::BTreeSet::new();
+				$(
+					<$provider as $crate::DataProviderExtended<$key, $timestamped_value>>::get_all_values()
+						.into_iter()
+						.for_each(|(k, _)| { keys.insert(k); });
```
xlc, line 67: not sure if `.map(|(k, _) k|).collect::<BTreeSet>::()` will work

xlc (approved):

traits/src/data_provider.rs
shaopengw: The `if let` below are to filter out `None` values, or we have to do a `retain`.

traits/src/data_provider.rs
shaopengw: Hmm, I tried, it works but doesn't read as straightforward as the previous way IMO, and also had to introduce an extra vec:

```rust
let keys = vec![$( <$provider as $crate::DataProviderExtended<$key, $timestamped_value>>::get_all_values() ),*]
    .into_iter()
    .flatten()
    .map(|(k, _)| k).collect::<sp_std::collections::btree_set::BTreeSet<_>>();
```

PR #261, opened 2020-08-25
set weight by WeightInfo

### LINKED ISSUE #260: use `WeightInfo` to overwrite weights for auction and currencies module

wangjj9219:
close #260 

set weight by `WeightInfo` for these not completely independent(depend on other handlers) modules so that runtime is able to overwrite the weight


shawntabrizi:
@wangjj9219 do you have any feedback about this format for weights? Better right?

xlc (approved):

shaopengw (approved):

Issue #247, opened 2020-08-12
CurrencyAdapter


xlc:
We need a `struct CurrencyAdapter<BasicCurrency>` that `impl Currency` so we can make currency from `orml-tokens` compatible with pallets from Substrate, such as treasury & elections-phragmen

PR #262, opened 2020-08-26
Add CurrencyAdapter

zjb0807:
Closes: #247

### DIFF
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -43,7 +43,9 @@ use frame_support::{
 };
 use frame_system::ensure_signed;
 use sp_runtime::{
-	traits::{AtLeast32Bit, CheckedAdd, CheckedSub, MaybeSerializeDeserialize, Member, Saturating, StaticLookup, Zero},
+	traits::{
+		AtLeast32BitUnsigned, CheckedAdd, CheckedSub, MaybeSerializeDeserialize, Member, Saturating, StaticLookup, Zero,
+	},
 	DispatchError, DispatchResult, RuntimeDebug,
 };
 use sp_std::{
@@ -68,7 +70,7 @@ pub trait Trait: frame_system::Trait {
 	type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
 
 	/// The balance type
-	type Balance: Parameter + Member + AtLeast32Bit + Default + Copy + MaybeSerializeDeserialize;
+	type Balance: Parameter + Member + AtLeast32BitUnsigned + Default + Copy + MaybeSerializeDeserialize;
 
 	/// The amount type, should be signed version of `Balance`
 	type Amount: Signed
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -45,7 +45,7 @@ use frame_support::{
 		LockableCurrency as PalletLockableCurrency, ReservableCurrency as PalletReservableCurrency, SignedImbalance,
 		WithdrawReasons,
 	},
-	weights::constants::WEIGHT_PER_MICROS,
+	weights::Weight,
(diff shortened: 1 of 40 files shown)


tokens/src/lib.rs
```diff
@@ -615,3 +626,217 @@ impl<T: Trait> MultiReservableCurrency<T::AccountId> for Module<T> {
...
+
+	fn deposit_into_existing(
+		who: &T::AccountId,
+		value: Self::Balance,
+	) -> result::Result<Self::PositiveImbalance, DispatchError> {
+		Module::<T>::deposit(GetCurrencyId::get(), who, value).map(|_| Self::PositiveImbalance::new(value))
+	}
+
+	fn deposit_creating(who: &T::AccountId, value: Self::Balance) -> Self::PositiveImbalance {
+		Module::<T>::deposit(GetCurrencyId::get(), who, value).map_or_else(
```
xlc, line 743: Does `Module::<T>::deposit` modify `TotalIssuance`? Make sure it use the one that doesn't modify `TotalIssuance`.

xlc (approved):
Should have some more tests for imbalance creation (via withdraw / deposit) and non-trivial Currency method implementation.

tokens/src/lib.rs
```diff
@@ -416,7 +430,7 @@ impl<T: Trait> MultiCurrency<T::AccountId> for Module<T> {
 	/// or `can_slash` wasn't used appropriately.
 	fn slash(currency_id: Self::CurrencyId, who: &T::AccountId, amount: Self::Balance) -> Self::Balance {
 		if amount.is_zero() {
-			return amount;
+			return Zero::zero();
```
shaopengw, line 433: Hmm, why creating a new `Zero` value instead of just return `amount` as zero?

tokens/src/mock.rs
```diff
@@ -36,7 +45,7 @@ parameter_types! {
 	pub const AvailableBlockRatio: Perbill = Perbill::one();
 }
 
-type AccountId = u128;
+type AccountId = u64;
```
shaopengw, line 48: Why changing it to `u64`?

tokens/src/mock.rs
```diff
@@ -89,6 +98,186 @@ impl OnReceived<AccountId, CurrencyId, Balance> for MockOnReceived {
...
+	type CandidacyBond = CandidacyBond;
+	type VotingBond = VotingBond;
+	type TermDuration = TermDuration;
+	type DesiredMembers = DesiredMembers;
+	type DesiredRunnersUp = DesiredRunnersUp;
+	type LoserCandidate = ();
+	type KickedMember = ();
+	type BadReport = ();
+	type WeightInfo = ();
+}
```
shaopengw, line 279: Maybe I missed something, but didn't see any usage for the new added mocks.

shaopengw (commented):
Some nitpicks. And could you add a unit test for the pallet currency adapter, just a basic transfer should be fine.

tokens/src/mock.rs
zjb0807: Because `pallet_elections_phragmen` and `pallet_treasury` module dispatch is private function. Maybe I can test it from genesis config.

tokens/src/mock.rs
zjb0807:
```
impl Contains<u64> for TenToFourteen {
```
Other defines are u64, it is more convenient to modify here

tokens/src/lib.rs
zjb0807: https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L1037

Refer here. You are right, I'll recover it.

shaopengw (approved):

PR #280, opened 2020-09-10
support storage map start with random position

wangjj9219:
because Substrate didn't expose some necessary structs and traits, I copy some implementation for StorageMap/StorageDoubleMap iterator

### DIFF
--- a/utilities/Cargo.toml
+++ b/utilities/Cargo.toml
@@ -13,13 +13,13 @@ codec = { package = "parity-scale-codec", version = "1.3.0", default-features = 
 frame-support = { version = "2.0.0-rc6", default-features = false }
 sp-runtime = { version = "2.0.0-rc6", default-features = false }
 sp-std = { version = "2.0.0-rc6", default-features = false }
+sp-io = { version = "2.0.0-rc6" }
 
 integer-sqrt = "=0.1.3" # https://github.com/derekdreery/integer-sqrt-rs/issues/12
 
 [dev-dependencies]
 serde_json = "1.0.53"
 frame-system = { version = "2.0.0-rc6" }
-sp-io = { version = "2.0.0-rc6" }
 
 clear_on_drop = { version = "0.2.4", features = ["no_cc"] }	# https://github.com/paritytech/substrate/issues/4179
 
@@ -31,4 +31,5 @@ std = [
 	"frame-support/std",
 	"sp-runtime/std",
 	"sp-std/std",
+	"sp-io/std",
 ]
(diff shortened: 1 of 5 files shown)


xlc (commented):
Instead of have a random iterator, should be an iterator start from a given prefix. 

And then we can have another random prefix generator.

The idea is we do random walk only on first run, and rest will continue from the last key.

xlc:
Should create issues on Substrate to request to make necessary structs and traits public.

utilities/src/offchain_worker/iterator_helper.rs
```diff
@@ -61,89 +215,144 @@ impl<T> Iterator for MapIterator<T> {
...
-	StorageDoubleMap<K1, K2, V>
-{
-	/// The type that iterates over all `(key2, value)`.
-	type PrefixIterator: Iterator<Item = (K2, V)>;
+/// Shim for MapIterator, add more features
+pub struct MapIteratorShim<'a, 'b, T, L: Lockable> {
+	pub map_iterator: MapIterator<T>,
+	pub remain_iterator_count: Option<u32>,
+	pub lock_guard: Option<StorageLockGuard<'a, 'b, L>>,
+	pub unfinished: bool,
```
xlc, line 223: make it `finished` to avoid double negative which are confusing

utilities/src/offchain_worker/iterator_helper.rs
```diff
@@ -61,89 +215,144 @@ impl<T> Iterator for MapIterator<T> {
...
-pub trait RandomIterableStorageDoubleMap<K1: FullCodec, K2: FullCodec, V: FullCodec>:
-	StorageDoubleMap<K1, K2, V>
-{
-	/// The type that iterates over all `(key2, value)`.
-	type PrefixIterator: Iterator<Item = (K2, V)>;
+/// Shim for MapIterator, add more features
+pub struct MapIteratorShim<'a, 'b, T, L: Lockable> {
+	pub map_iterator: MapIterator<T>,
+	pub remain_iterator_count: Option<u32>,
+	pub lock_guard: Option<StorageLockGuard<'a, 'b, L>>,
```
xlc, line 222: do we need to pass this? can we do the extend lock in the code that using this iterator?
this prevent us using this iterator within wasm runtime. we don't have a use case yet but I don't see why we need to prevent this possibility.

utilities/Cargo.toml
```diff
@@ -13,13 +13,13 @@ codec = { package = "parity-scale-codec", version = "1.3.0", default-features =
 frame-support = { version = "2.0.0-rc6", default-features = false }
 sp-runtime = { version = "2.0.0-rc6", default-features = false }
 sp-std = { version = "2.0.0-rc6", default-features = false }
+sp-io = { version = "2.0.0-rc6" }
```
xlc, line 16:
```suggestion
sp-io = { version = "2.0.0-rc6", default-features = false }
```

xlc (approved):

PR #281, opened 2020-09-11
Adjust claim-rewards

zjb0807:


rewards/src/lib.rs
```diff
@@ -172,12 +172,14 @@ impl<T: Trait> Module<T> {
 					return;
 				}
 
+				// pay reward to `who`
```
xlc, line 175: can also make `Pools::<T>::mutate(pool, |pool_info| {` to be `try_mutate` and just return the payout result

xlc (approved):

zjb0807:
To prevent users withdraw  from dex and share is not reduced, close PR

PR #304, opened 2020-10-13
Add GetByKey trait and get_by_key_type macro.

shaopengw:


traits/src/lib.rs
```diff
@@ -16,13 +16,15 @@ pub use currency::{
 	LockIdentifier, MultiCurrency, MultiCurrencyExtended, MultiLockableCurrency, MultiReservableCurrency, OnReceived,
 };
 pub use data_provider::{DataFeeder, DataProvider, DataProviderExtended};
+pub use get_by_key::GetByKey;
```
xlc, line 19: Do you need to expose `get_by_key_type` macro here?

xlc (approved):
Some more rustdocs will be helpful

traits/src/lib.rs
shaopengw: `#[macro_export]` exports the macro at crate root.

xlc (approved):

PR #325, opened 2020-11-25
refactor tokens

### LINKED ISSUE #324: orml-tokens: ExistentialDeposit
- orml-traits
  - trait MultiCurrency
    - fn minimum_balance(currency_id: CurrencyId)
- orml-tokens
  - Trait
    - type ExistentialDeposit: GetByKey<CurrencyId, Balance>
    - type OnDust
      - dust should send to treausry account
  - Module
    - update transfer to handle existential deposit requirement, use pallet-balances as reference
    - update system ref count when balances changed from zero or to zero

Relates: https://github.com/paritytech/substrate/issues/7563

wangjj9219:
close #324 

1. handle dust for these accounts which total is below `ExistantialDeposit` except for module accounts(treat module account as system account).
2. use `mutate_exist` to replace `mutate` to update AccountData storage so that remove those account whose total is 0.
3. Increase/decrease ref count when adding/removing AccountData storage.

### DIFF
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -49,11 +49,10 @@ use frame_support::{
 	Parameter, StorageMap,
 };
 use frame_system::ensure_signed;
-use sp_core::{crypto::AccountId32, TypeId};
 use sp_runtime::{
 	traits::{
-		AtLeast32BitUnsigned, Bounded, CheckedAdd, CheckedSub, MaybeSerializeDeserialize, Member, Saturating,
-		StaticLookup, Zero,
+		AccountIdConversion, AtLeast32BitUnsigned, Bounded, CheckedAdd, CheckedSub, MaybeSerializeDeserialize, Member,
+		Saturating, StaticLookup, Zero,
 	},
 	DispatchError, DispatchResult, ModuleId, RuntimeDebug,
 };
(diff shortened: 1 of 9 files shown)


tokens/src/lib.rs
```diff
@@ -288,29 +306,107 @@ decl_error! {
 }
 
 impl<T: Trait> Module<T> {
+	/// Check whether account_id is a module account
+	fn is_module_account_id(account_id: T::AccountId) -> bool {
+		let account: AccountId32 = Into::<AccountId32>::into(Into::<T::AccountIdConvert>::into(account_id));
```
xlc, line 311: Could use `AccountIdConversion::try_from_account(account_id).is_some()`

tokens/src/lib.rs
```diff
@@ -288,29 +306,107 @@ decl_error! {
...
+		who: &T::AccountId,
+		currency_id: T::CurrencyId,
+		new: AccountData<T::Balance>,
+		existed: bool,
+	) -> Option<AccountData<T::Balance>> {
+		let total = new.total();
+		if total < Self::existential_deposit(currency_id) && !Self::is_module_account_id(who.clone()) {
+			// remove dust
+			if !total.is_zero() {
+				TotalIssuance::<T>::mutate(currency_id, |t| {
```
xlc, line 334: Shouldn't modify `TotalIssuance` here. It is up to OnDust to decide if we want to burn the token (modify `TotalIssuance` or transfer to treasury or doing something else.

xlc (changes requested):

tokens/src/lib.rs
wangjj9219: `TotalInsurance` is still deducted here to avoid interface refactoring

xlc (approved):

shaopengw (approved):

PR #329, opened 2020-11-30
Add public minting feature

str-mv:
Public minting feature

### DIFF
--- a/nft/src/lib.rs
+++ b/nft/src/lib.rs
@@ -142,7 +142,7 @@ impl<T: Trait> Module<T> {
 			Ok(current_id)
 		})?;
 
-		let mode: MintMode = if public_mint == true {
+		let mode: MintMode = if public_mint {
 			MintMode::Public
 		} else {
 			MintMode::Private
@@ -226,7 +226,7 @@ impl<T: Trait> Module<T> {
 			let mut info = class_info.as_mut().ok_or(Error::<T>::ClassNotFound)?;
 			ensure!(info.owner == *owner, Error::<T>::NoPermission);
 
-			let mode: MintMode = if public == true {
+			let mode: MintMode = if public {
 				MintMode::Public
 			} else {
 				MintMode::Private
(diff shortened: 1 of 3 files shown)


nft/src/lib.rs
```diff
@@ -171,10 +191,15 @@ impl<T: Trait> Module<T> {
 	) -> Result<T::TokenId, DispatchError> {
 		NextTokenId::<T>::try_mutate(class_id, |id| -> Result<T::TokenId, DispatchError> {
 			let token_id = *id;
-			*id = id.checked_add(&One::one()).ok_or(Error::<T>::NoAvailableTokenId)?;
-
 			Classes::<T>::try_mutate(class_id, |class_info| -> DispatchResult {
 				let info = class_info.as_mut().ok_or(Error::<T>::ClassNotFound)?;
+
+				ensure!(
+					info.mint_mode == MintMode::Public || (info.mint_mode == MintMode::Private && info.owner == *owner),
```
usetech-llc, line 198: This is probably extraneous condition check? Can it be replaced with 
`info.mint_mode == MintMode::Public || info.owner == *owner`?

gregzaitsev, line 198: This is probably extraneous condition check? Can it be replaced with
`info.mint_mode == MintMode::Public || info.owner == *owner`?

str-mv: That's a good question.
I prefer the concept to write stateless checks. In this case, enumeration state.
For example, if we added the Invalid option, then in the existing version it would not work

gregzaitsev: Ok, makes sense, thanks!

nft/src/lib.rs
```diff
@@ -40,6 +48,8 @@ pub struct ClassInfo<TokenId, AccountId, Data> {
 	pub total_issuance: TokenId,
 	/// Class owner
 	pub owner: AccountId,
+	/// Mint mode
+	pub mint_mode: MintMode,
```
xlc, line 52: why not just `allow_public_mint: bool`? the public API is passing `public: bool` anyway

xlc (changes requested):
I don't see why this is needed here.

I can see why you will want this feature, but it should be implemented in your custom logic pallet instead of orml-nft.

The `mint_mode` can be stored in the class data and you can do your permission check before calling `mint`.

I intentionally to make sure there are no permission check handling logic inside this pallet because everyone will have different requirements and it just won't be possible to support all of them.

PR #332, opened 2020-12-02
Integrate Rococo V1

shaunxw:


shaopengw:
WIP, made ready for review to trigger CI.

olanod:
Looking nice :) Wondering if there are mayor blockers like protocol changes that stop this from being merged? would you say is safe enough to to try out xtokens in its current state even if its just locally?

xlc:
You can definitely give it a try. One reason we have not merged this yet is this branch is targeting to rococo-v1, which is a fast moving target and have breaking changes from time to time.

PR #351, opened 2021-01-08
Add genesis config for nft

pmensik:
The initial idea for a genesis config of the NFT pallet. The config accepts `BTreeMap<T::AccountId, BTreeMap<(Vec<u8>, T::ClassData), Vec<(Vec<u8>, T::TokenData)>>>` that basically says each account has a map where the key is a vector of tuples for token class data (metadata and ClassData) with the value of vector of tokens to be minted. 
However, I can't get this code built, so I will appreciate any help with finishing this PR

```
   Compiling orml-nft v0.3.5-dev (/home/pmensik/work/polkadot/open-runtime-module-library/nft)
error[E0463]: can't find crate for `serde`
   --> src/lib.rs:100:1
    |
100 | / decl_storage! {
101 | |     trait Store for Module<T: Config> as NonFungibleToken {
102 | |         /// Next available class ID.
103 | |         pub NextClassId get(fn next_class_id): T::ClassId;
...   |
161 | |     }
162 | | }
    | |_^ can't find crate
    |
```

### DIFF
--- a/nft/Cargo.toml
+++ b/nft/Cargo.toml
@@ -8,20 +8,22 @@ authors = ["Acala Developers"]
 edition = "2018"
 
 [dependencies]
+serde = { version = "1.0.111", optional = true }
 codec = { package = "parity-scale-codec", version = "1.3.0", default-features = false }
-sp-std = {  version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
-sp-runtime = { version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
+sp-std = {  version = "2.0.0", default-features = false }
+sp-runtime = { version = "2.0.0", default-features = false }
 
-frame-support = { version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
-frame-system = { version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
+frame-support = { version = "2.0.0", default-features = false }
+frame-system = { version = "2.0.0", default-features = false }
 
 [dev-dependencies]
-sp-io = {  version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
-sp-core = { version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
+sp-io = {  version = "2.0.0", default-features = false }
+sp-core = { version = "2.0.0", default-features = false }
 
 [features]
 default = ["std"]
 std = [
+	"serde",
 	"codec/std",
 	"sp-std/std",
 	"sp-runtime/std",
--- a/nft/Cargo.toml
+++ b/nft/Cargo.toml
@@ -10,14 +10,14 @@ edition = "2018"
 [dependencies]
 serde = { version = "1.0.111", optional = true }
 codec = { package = "parity-scale-codec", version = "1.3.0", default-features = false }
-sp-std = {  version = "2.0.0", default-features = false }
+sp-std = { version = "2.0.0", default-features = false }
 sp-runtime = { version = "2.0.0", default-features = false }
 
 frame-support = { version = "2.0.0", default-features = false }
 frame-system = { version = "2.0.0", default-features = false }
 
 [dev-dependencies]
-sp-io = {  version = "2.0.0", default-features = false }
+sp-io = { version = "2.0.0", default-features = false }
 sp-core = { version = "2.0.0", default-features = false }
 
 [features]
--- a/nft/Cargo.toml
+++ b/nft/Cargo.toml
@@ -9,15 +9,15 @@ edition = "2018"
 
 [dependencies]
 codec = { package = "parity-scale-codec", version = "1.3.0", default-features = false }
-sp-std = { version = "2.0.0", default-features = false }
-sp-runtime = { version = "2.0.0", default-features = false }
+sp-std = {  version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
+sp-runtime = { version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
 
-frame-support = { version = "2.0.0", default-features = false }
-frame-system = { version = "2.0.0", default-features = false }
+frame-support = { version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
+frame-system = { version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
 
 [dev-dependencies]
-sp-io = { version = "2.0.0", default-features = false }
-sp-core = { version = "2.0.0", default-features = false }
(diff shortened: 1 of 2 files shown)


nft/Cargo.toml
```diff
@@ -9,15 +9,15 @@ edition = "2018"
 
 [dependencies]
 codec = { package = "parity-scale-codec", version = "1.3.0", default-features = false }
-sp-std = { version = "2.0.0", default-features = false }
-sp-runtime = { version = "2.0.0", default-features = false }
+sp-std = {  version = "2.0.0", default-features = false, git = "https://github.com/paritytech/substrate", rev = "b2750359" }
```
xlc, line 12: please revert chose change and update Cargo.dev.toml instead.

I've already bumped Substrate version at #350 so no need to bump again

nft/src/lib.rs
```diff
@@ -91,6 +91,12 @@ pub type ClassInfoOf<T> =
 	ClassInfo<<T as Config>::TokenId, <T as frame_system::Config>::AccountId, <T as Config>::ClassData>;
 pub type TokenInfoOf<T> = TokenInfo<<T as frame_system::Config>::AccountId, <T as Config>::TokenData>;
 
+#[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug)]
+pub struct NftGenesisConfig<AccountId, ClassData, TokenData> {
```
xlc: this isn't used at all?

nft/src/lib.rs
```diff
@@ -109,6 +115,50 @@ decl_storage! {
 		#[cfg(not(feature = "disable-tokens-by-owner"))]
 		pub TokensByOwner get(fn tokens_by_owner): double_map hasher(twox_64_concat) T::AccountId, hasher(twox_64_concat) (T::ClassId, T::TokenId) => Option<()>;
 	}
+	add_extra_genesis {
+		config(endowed_accounts): BTreeMap<T::AccountId, BTreeMap<(Vec<u8>, T::ClassData), Vec<(Vec<u8>, T::TokenData)>>>;
```
xlc, line 113: this type is super confusing... you can use type alias to improve it. something like

```
type GenesisTokenData<T> = (Vec<u8>, <T as Trait>::TokenData);
type GenesisClassData<T> = (Vec<u8>, <T as Trait>::ClassData);)
```

Also instead of `AccountId => (ClassData => TokenData)`, I will prefer `ClassData => (AccountId, TokenData)`

xlc:
You need those to fix the serde error

https://github.com/open-web3-stack/open-runtime-module-library/blob/2de1e024781a7744a3aeb82229ebcb555163452a/auction/Cargo.toml#L11

https://github.com/open-web3-stack/open-runtime-module-library/blob/2de1e024781a7744a3aeb82229ebcb555163452a/auction/Cargo.toml#L28

pmensik:
@xlc I've updated the solution and changed a type according to your comment (I've just added an `AccountId` to a `GenesisClassData` because class data needs an owner as well). Thanks for your review!

nft/Cargo.toml
```diff
@@ -8,20 +8,22 @@ authors = ["Acala Developers"]
 edition = "2018"
 
 [dependencies]
+serde = { version = "1.0.111", optional = true }
 codec = { package = "parity-scale-codec", version = "1.3.0", default-features = false }
-sp-std = { version = "2.0.0", default-features = false }
+sp-std = {  version = "2.0.0", default-features = false }
```
xlc, line 13:
```suggestion
sp-std = { version = "2.0.0", default-features = false }
```

nft/Cargo.toml
```diff
@@ -8,20 +8,22 @@ authors = ["Acala Developers"]
...
 sp-runtime = { version = "2.0.0", default-features = false }
 
 frame-support = { version = "2.0.0", default-features = false }
 frame-system = { version = "2.0.0", default-features = false }
 
 [dev-dependencies]
-sp-io = { version = "2.0.0", default-features = false }
+sp-io = {  version = "2.0.0", default-features = false }
```
xlc, line 20:
```suggestion
sp-io = { version = "2.0.0", default-features = false }
```

nft/src/lib.rs
```diff
@@ -61,7 +61,7 @@ pub trait Config: frame_system::Config {
 	/// The token ID type
 	type TokenId: Parameter + Member + AtLeast32BitUnsigned + Default + Copy;
 	/// The class properties type
-	type ClassData: Parameter + Member;
+	type ClassData: Parameter + Member + Ord;
```
xlc, line 64: I don't like this additional trait bound just so we can have it in genesis.

One way to avoid it is just use `Vec` instead of `BTreeMap`

nft/src/lib.rs
```diff
@@ -109,6 +116,28 @@ decl_storage! {
 		#[cfg(not(feature = "disable-tokens-by-owner"))]
 		pub TokensByOwner get(fn tokens_by_owner): double_map hasher(twox_64_concat) T::AccountId, hasher(twox_64_concat) (T::ClassId, T::TokenId) => Option<()>;
 	}
+	add_extra_genesis {
+		config(endowed_accounts): BTreeMap<GenesisClassData<T>, BTreeMap<T::AccountId, Vec<GenesisTokenData<T>>>>;
+
+		build(|config: &GenesisConfig<T>| {
+			config.endowed_accounts.iter().for_each(|(class_data, tokens)| {
+				let result = Module::<T>::create_class(&class_data.0, class_data.1.to_vec(), class_data.2.clone());
```
xlc, line 124:
```suggestion
				let class_id = Module::<T>::create_class(&class_data.0, class_data.1.to_vec(), class_data.2.clone()).expect("Create class cannot fail while building genesis");
```

Never ignore error silencely

nft/Cargo.toml
pmensik: Fixed, thanks for the note

nft/src/lib.rs
pmensik: Thank you, I didn't think about `expect`

nft/src/lib.rs
pmensik: @xlc Hmm, it's probably not a good idea to restrict the type furthermore. However, I feel that the map is a really good fit in this situation (because it's a mapping of token classes to tokens themselves), is it possible to use another implementation that doesn't require `Ord`? Like this [one](https://substrate.dev/rustdocs/v2.0.0/hashbrown/struct.HashMap.html). On the other hand, introducing an external dependency doesn't sound like a good idea either. So something like this feel alright?

```
// each token will have an owner
pub type GenesisTokenData<T> = (T::AccountId, Vec<u8>, <T as Config>::TokenData);
// each token class will contain its tokens
pub type GenesisTokens<T> = (
	<T as frame_system::Config>::AccountId,
	Vec<u8>,
	<T as Config>::ClassData,
	Vec<GenesisTokenData<T>>
);
config(tokens): Vec<GenesisTokens<T>> {}
```

xlc: Yep that should work.

pmensik: Fixed, I just wonder whether should I add any documentation for the types (in the form comments). Or is self-explanatory enough?

xlc: some simple comment will be helpful, especially for `Vec<u8>` part

xlc:
Please merge master to fix CI

nft/src/lib.rs
pmensik: Done

xlc (approved):
Thanks for the contribution!

PR #355, opened 2021-01-14
Migrate to use pallet!

wangjj9219:


auction/src/lib.rs
```diff
@@ -139,113 +106,113 @@ decl_module! {
...
+			if let Some(auction) = Auctions::<T>::take(&id) {
+				if let Some(end_block) = auction.end {
+					AuctionEndTime::<T>::remove(end_block, id);
+				}
+			}
 		}
+	}
 
-		Ok(auction_id)
+	#[pallet::error]
```
xlc, line 186: I think it is better to move Error / Event / Storages to before all the methods

xlc (approved):
If it compiles and tests are passing, I think it should be good.

shaopengw (approved):
LGTM

PR #361, opened 2021-02-03
use account provider

### LINKED ISSUE #360: use account providers for tokens module
Now, reference counting of provider and consumerr decides the account existence, :https://github.com/paritytech/substrate/commit/7a79f54a5d92cecba1d9c1e4da71df1e8a6ed91b

update substrate and use new account existence for related modules

wangjj9219:
close #360

### DIFF
--- a/Cargo.dev.toml
+++ b/Cargo.dev.toml
@@ -18,21 +18,21 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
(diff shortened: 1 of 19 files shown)


wangjj9219:
The new version of the `pallet-elections-phragmen` caused some changes in the test case result, may be we should move these related test cases into integrate tests.

xlc (approved):

xlc:
We want to test orml-tokens works with pallet-elections-phragmen so I think it is fine to have the test here, unless it is becoming too painful to maintain.

PR #379, opened 2021-02-22
Add `transfer_all` to Currencies

mzxyz:
**Changes**:
- Add `transfer_all` to currencies
- Add test for transfer_all

Not fully understand the requirements for # 375, for `tokens` module there has already have `transfer_all` function.

**Further steps**:
- [ ] Need to generate correct weight for `transfer_all_non_native_currency`

### DIFF
--- a/currencies/src/lib.rs
+++ b/currencies/src/lib.rs
@@ -77,6 +77,7 @@ pub mod module {
 
 	pub trait WeightInfo {
 		fn transfer_non_native_currency() -> Weight;
+		fn transfer_all_non_native_currency() -> Weight;
 		fn transfer_native_currency() -> Weight;
 		fn update_balance_non_native_currency() -> Weight;
 		fn update_balance_native_currency_creating() -> Weight;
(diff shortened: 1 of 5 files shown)


currencies/src/lib.rs
```diff
@@ -156,6 +157,24 @@ pub mod module {
...
+		/// transactor.
+		#[pallet::weight(T::WeightInfo::transfer_all_non_native_currency())]
+		pub fn transfer_all(
+			origin: OriginFor<T>,
+			dest: <T::Lookup as StaticLookup>::Source,
+			currency_id: CurrencyIdOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let from = ensure_signed(origin)?;
+			let to = T::Lookup::lookup(dest)?;
+			let balance = <Self as MultiCurrency<T::AccountId>>::free_balance(currency_id, &from);
```
xlc, line 172: this needs to be free - frozen

xlc (changes requested):

currencies/src/lib.rs
HZHAIX: Do you mean apply this for `transfer_all` in token module as we only define it in `orml-tokens`

xlc:
I think the best way is add `transfer_all` to https://github.com/open-web3-stack/open-runtime-module-library/blob/master/traits/src/currency.rs and use it in orml-currencies

Also update the tests to check the case when user have some locked token.

HZHAIX:
> I think the best way is add `transfer_all` to https://github.com/open-web3-stack/open-runtime-module-library/blob/master/traits/src/currency.rs and use it in orml-currencies
> 
> Also update the tests to check the case when user have some locked token.

Make sense

Issue #387, opened 2021-03-03
Fix CI on master


xlc:
https://github.com/open-web3-stack/open-runtime-module-library/pull/386/checks?check_run_id=2023422607

Looks like new release of frame benchmarking breaks master CI

PR #389, opened 2021-03-05
Fix ci

zjb0807:
Closes: #387


xlc (approved):
Interesting this fixes the issue...

PR #394, opened 2021-03-11
Polkadex <> ORML

Gauthamastro:
I would like to add Polkadex project as we are using ORML pallets in Polkadex Project


Gauthamastro:
@xlc Can you please add this PR too. I think the PR got missed in the buzz :)

Issue #340, opened 2020-12-16
Weight Meter


xlc:
https://hackmd.io/@xlc/r17SnD3oD

PR #395, opened 2021-03-11
weight_meter

ermalkaleci:
#340


weight-meter/weight-meter-procedural/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+use syn::{Attribute, FnArg, Ident, ImplItem, ImplItemMethod, Item, ItemFn, ItemImpl, ItemMod, Pat};
+
+#[proc_macro_attribute]
+pub fn start_with(attr: TokenStream, item: TokenStream) -> TokenStream {
+	let base_weight: syn::Expr = syn::parse(attr).unwrap();
+	let ItemFn { attrs, vis, sig, block } = syn::parse(item).unwrap();
+	(quote! {
+		#(#attrs)*
+		#vis #sig {
+			orml_weight_meter::start_with(#base_weight);
```
xlc, line 14: should avoid hardcode crate name in macro. should be something like `$crate::orml_weight_meter::start_with`

weight-meter/weight-meter-procedural/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+			orml_weight_meter::using(#weight);
+			#block
+		}
+	})
+	.into()
+}
+
+#[proc_macro_attribute]
+pub fn method_benchmarks(_attr: TokenStream, input: TokenStream) -> TokenStream {
+	#[cfg(feature = "runtime-benchmarks")] // Inject methods if we're benchmarking
```
xlc, line 39: could just put all the code with runtime-benchmarks into a mod so we only need one feature gate

weight-meter/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::weights::Weight;
+
+struct Meter {
+	used_weight: Weight,
+	// Deep gets incremented when entering call or a sub-call
+	// This is used to avoid miscalculation during sub-calls
+	deep: usize,
```
xlc, line 9:
```suggestion
	depth: usize,
```

xlc: why it is usize?

weight-meter/README.md
```diff
@@ -0,0 +1,66 @@
+# Weight Meter
+
+## How to use WeightMeter
```
xlc, line 3: those can be in the rustdoc of lib.rs

weight-meter/src/lib.rs
ermalkaleci: No reason at all, u8 works as well. This is going to be 1 most of the time. `usize` was the easiest choice

weight-meter/weight-meter-procedural/src/lib.rs
ermalkaleci: yeah makes sense

ermalkaleci: Seems like this works only for macros. Procedural macros are unhygienic https://doc.rust-lang.org/reference/procedural-macros.html

weight-meter/src/lib.rs
xlc: Should avoid usize because the size is not fixed. So could go u32 here.

weight-meter/weight-meter-procedural/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+mod method_injector;
+
+#[proc_macro_attribute]
+pub fn start_with(attr: TokenStream, item: TokenStream) -> TokenStream {
+	let base_weight: syn::Expr = syn::parse(attr).unwrap();
+	let ItemFn { attrs, vis, sig, block } = syn::parse(item).unwrap();
+	(quote! {
+		#(#attrs)*
+		#vis #sig {
+			::orml_weight_meter::start_with(#base_weight);
```
xlc, line 14: orml_weight_meter could be imported with different name

xlc (changes requested):

weight-meter/weight-meter-procedural/src/lib.rs
xlc: This is how substrate solves it https://github.com/paritytech/substrate/blob/master/frame/support/procedural/tools/src/lib.rs

xlc:
replaced by #478

PR #401, opened 2021-03-14
Use xcm-handler to execute XCM locally

### LINKED ISSUE #399: Use xcm-handler to execute XCM locally
Now `xcm-handler` can execute XCM locally with an `origin` parameter (via [this PR](https://github.com/paritytech/cumulus/pull/357)), we don't have to use the `xcm-executor` workaround.

shaunxw:
Closes #399

### DIFF
--- a/xcm-support/src/lib.rs
+++ b/xcm-support/src/lib.rs
@@ -12,10 +12,14 @@ use sp_std::{
 	result,
 };
 
-use xcm::v0::{Error, Junction, MultiAsset, MultiLocation, Result};
+use xcm::v0::{Error, Junction, MultiAsset, MultiLocation, Result, Xcm};
 use xcm_executor::traits::{FilterAssetLocation, LocationConversion, MatchesFungible, NativeAsset, TransactAsset};
 
-use frame_support::{log, traits::Get};
+use frame_support::{dispatch::DispatchResult, log, traits::Get};
+
+pub trait XcmHandler<AccountId> {
+	fn execute_xcm(origin: AccountId, xcm: Xcm) -> DispatchResult;
+}
 
 pub trait CurrencyIdConversion<CurrencyId> {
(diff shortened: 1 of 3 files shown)


xcm-support/src/lib.rs
```diff
@@ -12,10 +12,14 @@ use sp_std::{
...
 };
 
-use xcm::v0::{Error, Junction, MultiAsset, MultiLocation, Result};
+use xcm::v0::{Error, Junction, MultiAsset, MultiLocation, Result, Xcm};
 use xcm_executor::traits::{FilterAssetLocation, LocationConversion, MatchesFungible, NativeAsset, TransactAsset};
 
-use frame_support::{log, traits::Get};
+use frame_support::{dispatch::DispatchResult, log, traits::Get};
+
+pub trait XcmHandler<AccountId> {
```
xlc, line 20: shouldn't this be in cumulus / polkadot?

xcm-support/src/lib.rs
```diff
@@ -12,10 +12,14 @@ use sp_std::{
...
+	fn execute_xcm(origin: AccountId, xcm: Xcm) -> DispatchResult;
```
xlc, line 21:
```suggestion
	fn execute_xcm(origin: &AccountId, xcm: Xcm) -> DispatchResult;
```

xcm-support/src/lib.rs
shaunxw: There is no usage of this trait in Cumulus, not sure if it's good to add it there. We can probably open a PR to add it later.

xcm-support/src/lib.rs
shaunxw: The `xcm-handler` requires `AccountId` param.

Making it as reference would require `Clone` bound on `AccountId` generic and do clones.

xlc (approved):

PR #409, opened 2021-03-18
Handle unknown assets in TransactAsset impl

shaunxw:


xcm-support/src/currency_adapter.rs
```diff
@@ -46,35 +64,48 @@ pub struct MultiCurrencyAdapter<MultiCurrency, Matcher, AccountIdConverter, Acco
...
-		let who = AccountIdConverter::from_location(location)
-			.ok_or_else(|| XcmError::from(Error::AccountIdConversionFailed))?;
-		let currency_id =
-			CurrencyIdConverter::from_asset(asset).ok_or_else(|| XcmError::from(Error::CurrencyIdConversionFailed))?;
-		let amount: MultiCurrency::Balance = Matcher::matches_fungible(&asset)
-			.ok_or_else(|| XcmError::from(Error::AssetNotFound))?
-			.saturated_into();
-		MultiCurrency::deposit(currency_id, &who, amount).map_err(|e| XcmError::FailedToTransactAsset(e.into()))?;
-		Ok(())
+		UnknownAsset::deposit(asset, location).or_else(|_| {
```
xlc, line 85: should try known first and then try unknown

xcm-support/src/lib.rs
```diff
@@ -111,3 +111,21 @@ where
...
+	/// Deposit unknown asset.
+	fn deposit(asset: &MultiAsset, to: &MultiLocation) -> DispatchResult;
+
+	/// Withdraw unknown asset.
+	fn withdraw(asset: &MultiAsset, from: &MultiLocation) -> DispatchResult;
+}
+
+impl UnknownAsset for () {
+	fn deposit(_asset: &MultiAsset, _to: &MultiLocation) -> DispatchResult {
+		Ok(())
```
xlc, line 126: I think default should return Err?

xlc (changes requested):

xcm-support/src/currency_adapter.rs
shaunxw: Aha yes my bad. I thought it the wrong way (in order to make it return known asset deposit error). Will update.

xcm-support/src/lib.rs
shaunxw: True, it makes more sense to return a `CannotHandleUnknownAsset` error.

xlc (approved):

PR #437, opened 2021-04-06
Use CurrencyId convert.

shaunxw:


xcm-support/src/currency_adapter.rs
```diff
@@ -38,35 +37,53 @@ impl From<Error> for XcmError {
 ///
 /// If the asset is known, deposit/withdraw will be handled by `MultiCurrency`,
 /// else by `UnknownAsset` if unknown.
-pub struct MultiCurrencyAdapter<MultiCurrency, UnknownAsset, Matcher, AccountIdConverter, AccountId, CurrencyId>(
+pub struct MultiCurrencyAdapter<
+	MultiCurrency,
+	UnknownAsset,
+	Matcher,
+	AccountIdConvert,
```
xlc, line 44: may be `AccountId, AccountIdCovert, CurrencyId, CurrencyIdConvert` to make it more consistent

xlc (approved):

xcm-support/src/currency_adapter.rs
shaunxw: Updated.

PR #446, opened 2021-04-12
CLI Options

brettkolodny:
Added CLI options for specifying an output, local template, and local header file.

### DIFF
--- a/weight-gen/src/main.rs
+++ b/weight-gen/src/main.rs
@@ -1,0 +1,185 @@
+use serde::{Serialize, Deserialize};
+use clap::{AppSettings, Clap};
+use std::io::Read;
+
+#[derive(Clap)]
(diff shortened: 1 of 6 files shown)


weight-gen/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+  };
+
+  // Write benchmark to file or print to terminal if output path is not given.
+  if let Some(path) = opts.out {
+    let mut output_file = ::std::fs::File::create(&path)
+			.expect(&format!("Could not create output file at {}", &path));
+
+		handlebars
+			.render_template_to_write(&template, &hbs_data, &mut output_file)
+			.unwrap();
```
xlc, line 177: use expect instead of unwrap

xlc (approved):

PR #464, opened 2021-04-27
add whitelist_account

xlc:


shawntabrizi:
is there a reason you have forked benchmarking completely? We are changing it too much?

xlc:
Mainly most of pallets have tight coupling and it is impossible to write generic benchmark tests. So many of our benchmarks are ends up in the runtime https://github.com/AcalaNetwork/Acala/blob/master/runtime/mandala/src/benchmarking/cdp_engine.rs

PR #469, opened 2021-04-29
Remove common parameters from benchmarking.

shaunxw:


xlc (approved):

benchmarking/src/lib.rs
ljuanto, line 1: 25f99272fa881e2c7c63aea5e295d93ee91cede4

PR #478, opened 2021-05-05
New weight meter PR

ermalkaleci:


xlc (approved):

shaunxw (commented):
I think we need some unit tests like in `frame-support` to make sure all the macros expanded as expected.

Prints should be replaced with `log`.

weight-meter/src/lib.rs
```diff
@@ -2,6 +2,9 @@
 
 use frame_support::weights::Weight;
 
+// For use in mock file
+extern crate self as orml_weight_meter;
```
xlc, line 6: interesting you can do this why is it needed here? how about `use crate as orml_weight_meter` in mock.rs?

xlc (approved):

weight-meter/src/lib.rs
brettkolodny: That's what I had originally tried but for some reason I couldn't get it to work.

weight-meter/src/meter_std.rs
```diff
@@ -0,0 +1,44 @@
...
+
+thread_local! {
+	static METER: RefCell<Meter> = RefCell::new(Meter {
+		used_weight: 0,
+		depth: 0,
+	});
+}
+
+/// Start weight meter with base weight
+pub fn start_with() {
```
ermalkaleci, line 15:
```suggestion
pub fn start() {
```

PR #484, opened 2021-05-10
weight meter tests init

ermalkaleci:


xlc:
what's the main purpose of this? we don't need ui test to assert it compiles. it is mostly useful to assert correct error message is generated by macros

ermalkaleci:
> what's the main purpose of this? we don't need ui test to assert it compiles. it is mostly useful to assert correct error message is generated by macros

Yeah, maybe we don't need this for weight-meter cause macros are very simple. Maybe we need this for bencher

xlc:
Why do we need this for bencher? The normal unit test will ensure it compiles, why do we need to test the error message when it failed to compile?

PR #487, opened 2021-05-13
Converted Vec storage to BoundedVec

brettkolodny:


tokens/src/lib.rs
```diff
@@ -444,7 +446,9 @@ impl<T: Config> Pallet<T> {
 				frame_system::Pallet::<T>::dec_consumers(who);
 			}
 		} else {
-			<Locks<T>>::insert(who, currency_id, locks);
+			let bounded_locks: BoundedVec<BalanceLock<T::Balance>, T::MaxLocks> =
+				locks.to_vec().try_into().expect("Too many locked balances");
```
xlc, line 450: only use expect for things that can never fail. should modify the method signature to make it return `DispatchResult` instead

tokens/src/mock.rs
```diff
@@ -207,6 +207,7 @@ parameter_type_with_key! {
 
 parameter_types! {
 	pub DustAccount: AccountId = PalletId(*b"orml/dst").into_account();
+	pub MaxLocks: u32 = 100_000;
```
xlc, line 210: mark this small and write test to check behavior when exceed the limit

xlc (changes requested):

tokens/src/lib.rs
```diff
@@ -661,7 +669,7 @@ impl<T: Config> MultiLockableCurrency<T::AccountId> for Pallet<T> {
 		if let Some(lock) = new_lock {
 			locks.push(lock)
 		}
-		Self::update_locks(currency_id, who, &locks[..]);
+		Self::update_locks(currency_id, who, &locks[..])?;
 		Ok(())
```
xlc, line 673:
```suggestion
		Self::update_locks(currency_id, who, &locks[..])
```

xlc (approved):

xlc (approved):

PR #494, opened 2021-05-18
Update dependencies

zjb0807:


shaunxw:
Why bumping substrate deps? The current ref is `polkadot-v0.9.1` which is recommended for Kusama parachain.

zjb0807:
> Why bumping substrate deps? The current ref is `polkadot-v0.9.1` which is recommended for Kusama parachain.

Be consistent with Acala's dependency version.
Will there be any impact if I only upgrade Acala's dependencies?

shaunxw:
Acala's dependencies shouldn't be bumped either, as the `sp-*` libs version is very sensitive, and the inconsistency with validators' version could lead to collation failure.

PR #507, opened 2021-05-30
Bencher update

ermalkaleci:
Refactor bencher to build and run benches without the need of a build script. Bencher will take care of ignoring methods that are already been benchmarked.


xlc (commented):
Is it possible to pull build_wasm from Substrate instead of copy & paste here?

ermalkaleci:
> Is it possible to pull build_wasm from Substrate instead of copy & paste here?

I had to modify the code. Best way will be to make a PR to build_wasm

xlc (approved):

PR #513, opened 2021-06-07
Changed OrderedSet from Vec to BoundedVec

brettkolodny:


xlc (commented):
Need more tests for out of bound insert

oracle/src/lib.rs
```diff
@@ -89,6 +89,8 @@ pub mod module {
 
 		/// Weight information for extrinsics in this module.
 		type WeightInfo: WeightInfo;
+
+		type MaxHasDispatchedSize: Get<u32>;
```
xlc, line 93: need a doc comment

utilities/src/ordered_set.rs
```diff
@@ -60,74 +58,100 @@ impl<T: Ord> OrderedSet<T> {
...
+	fn from(v: BoundedVec<T, S>) -> Self {
 		Self::from(v)
 	}
 }
 
+#[cfg(feature = "std")]
+impl<T, S> fmt::Debug for OrderedSet<T, S>
+where
+	T: fmt::Debug,
+	S: Get<u32> + fmt::Debug,
```
xlc, line 75: I don't think we need this line?

brettkolodny: I can't seem to derive `Debug` and the code won't compile without it since it's needed for `assert` in the tests

xlc: I mean is `S: Get<u32> + fmt::Debug,` required?

brettkolodny: You're right that wasn't needed

PR #514, opened 2021-06-08
Update xtokens dependencies.

shaunxw:


xlc:
on ice until polkadot-v0.9.4 is created on cumulus

PR #522, opened 2021-06-15
Check to account ExistentialDeposit

zjb0807:


xlc (approved):
I think this should be fine but better to have additional pair of eyes

shaunxw (approved):
LGTM, docstring of `do_transfer` needs update to match the changes.

PR #538, opened 2021-07-05
Add XcmTransfer trait

### LINKED ISSUE #537: XCM: make xtokens `do_` functions pub
executing XCM transfers are currently limited to dispatchable calls.
can the private `do_` be made `pub` so xtokens transfers can be executed directly from other pallets?

mattsse:
# Changes
- add `XcmTransfer` trait as initial abstraction over xtokens pallet to transfer native currencies and `MultiAsset`
- implement `XcmTransfer` with some considerations:
  - I left some minor duplications in the `XcmTransfer::transfer` and the `transfer` dispatchable call, because I didn't wrap it in `with_xcm_execution_transaction` because I wasn't sure whether the  `XcmTransfer` trait also should use `with_xcm_execution_transaction`?
  - the return type is also `XcmExecutionResult`, which drops the used weight from `Outcome`, should the trait rather return `Result<Outcome, DispatchError>`?

- Closes #537


xtokens/src/lib.rs
```diff
@@ -395,6 +398,39 @@ pub mod module {
...
+			who: T::AccountId,
+			currency_id: T::CurrencyId,
+			amount: T::Balance,
+			dest: MultiLocation,
+			dest_weight: Weight,
+		) -> XcmExecutionResult {
+			if amount == Zero::zero() {
+				return Ok(Option::None);
+			}
+			let id: MultiLocation = T::CurrencyIdConvert::convert(currency_id.clone())
```
xlc, line 413: why the clone here?

xtokens/src/lib.rs
```diff
@@ -395,6 +398,39 @@ pub mod module {
...
+
+	impl<T: Config> XcmTransfer<T::AccountId, T::Balance, T::CurrencyId> for Pallet<T> {
+		fn transfer(
+			who: T::AccountId,
+			currency_id: T::CurrencyId,
+			amount: T::Balance,
+			dest: MultiLocation,
+			dest_weight: Weight,
+		) -> XcmExecutionResult {
+			if amount == Zero::zero() {
```
xlc, line 410:
```suggestion
			if amount.is_zero() {
```

xtokens/src/lib.rs
```diff
@@ -395,6 +398,39 @@ pub mod module {
...
+				return Ok(Option::None);
```
xlc, line 411:
```suggestion
				return Ok(None);
```

xtokens/src/lib.rs
```diff
@@ -395,6 +398,39 @@ pub mod module {
...
+		}
+
+		fn transfer_multi_asset(
+			who: T::AccountId,
+			asset: MultiAsset,
+			dest: MultiLocation,
+			dest_weight: Weight,
+		) -> XcmExecutionResult {
+			if Self::is_zero_amount(&asset) {
+				return Ok(Option::None);
```
xlc, line 429:
```suggestion
				return Ok(None);
```

shaunxw (commented):
I would prefer to have the `XcmTransfer` impl wrapped in `with_xcm_execution_transaction`. The transfer *must* be transactional anyway, or users lost their fund if XCM sending failed. It's better to do it in a lower level here instead of doing it everywhere.

xtokens/src/lib.rs
mattsse: using `Option::None` explicitly here was necessary because `MultiAsset::None` is imported with the xcm prelude.

xtokens/src/lib.rs
mattsse: updated this everywhere.

xtokens/src/lib.rs
mattsse: was a leftover after copy pasting this. removed.

mattsse:
> I would prefer to have the `XcmTransfer` impl wrapped in `with_xcm_execution_transaction`. The transfer _must_ be transactional anyway, or users lost their fund if XCM sending failed. It's better to do it in a lower level here instead of doing it everywhere.

* `XcmTransfer` impl now uses `with_xcm_execution_transaction`
* `XcmTransfer` trait functions return the `Outcome` directly:
    * instead of matching the outcome in `do_transfer_multiasset`, the dispatchable calls use `ensure_complete`

xtokens/src/lib.rs
```diff
@@ -395,16 +397,57 @@ pub mod module {
...
 		let res = f();
 		match res {
-			Ok(ref err) if err.is_none() => TransactionOutcome::Commit(res),
+			Ok(ref outcome) => {
+				if let Outcome::Complete(_) = outcome {
+					TransactionOutcome::Commit(res)
+				} else {
+					TransactionOutcome::Rollback(res)
+				}
+			}
```
shaunxw, line 450:
```suggestion
			Ok(ref outcome) if outcome.ensure_complete().is_ok() => {
				TransactionOutcome::Commit(res)
			}
```

shaunxw (commented):
LGTM.

Though we shoud hold this PR for a while to wait for unit tests being added. I'm working on it and it should be ready soon.

xtokens/src/lib.rs
mattsse: `ensure_complete` takes ownership of `outcome`, so this suggestions would require a clone.

shaunxw: Yeah you're right. In that case the outer `match` could be replaced with `if let` to get rid of the repeat `TransactionOutcome::Rollback(res)`, like:

```rust
if let Ok(ref outcome) = res {
    if let Outcome::Complete(_) = outcome {
        return TransactionOutcome::Commit(res);
    }
}
TransactionOutcome::Rollback(res)
```

mattsse: simplified even a bit more

shaunxw:
The unit tests are ready now. @mattsse Could you please merge master branch?

mattsse:
> The unit tests are ready now. @mattsse Could you please merge master branch?

@shaunxw just merged master branch

shaunxw (approved):

PR #539, opened 2021-07-08
Add orml-xcm module.

shaunxw:
Need some end to end test for HOMA cross-chain transact.


shaunxw:
Note I put orml-xcm unit tests in orml-xtokens module for now, to avoid repeating that great amount of mock. Will do refactor later to make xcm related mocks easier.

xlc (approved):

PR #542, opened 2021-07-09
fix some calculation which use FixedU128 as multiper

wangjj9219:


xlc (approved):
Do we need more tests?

PR #555, opened 2021-07-16
fix deposit_creating and deposit_into_existing

wangjj9219:
should add some check according to the impl of `pallet-balances`: https://github.com/paritytech/substrate/blob/0c971e0e154efe7918cc687ab99d477901fb4219/frame/balances/src/lib.rs#L1490-L1535


zjb0807 (approved):
LGTM.
`set_free_balance` and `set_reserved_balance` are dangerous and can cause unexpected things.

shaunxw (approved):

xlc (approved):

PR #565, opened 2021-07-21
Refactor tokens

wangjj9219:
Refactor as follows:
1. unify the handling for ED for `MutilCurrency`(deposit, withdraw, transfer) between `orml-tokens` and the adapter for `pallet-balances` .
2. do some general abstract
3. add some calls to facilitate the separate use of `orml-tokens`


tokens/src/lib.rs
```diff
@@ -334,30 +352,152 @@ pub mod module {
...
+		/// 99% of the time you want [`transfer`] instead.
+		///
+		/// The dispatch origin for this call must be `Signed` by the
+		/// transactor.
+		///
+		/// - `dest`: The recipient of the transfer.
+		/// - `currency_id`: currency type.
+		/// - `amount`: free balance amount to tranfer.
+		#[pallet::weight(T::WeightInfo::transfer())]
+		pub fn transfer_keep_alive(
```
xlc, line 373: new extrinsics needs to be added at end to avoid change extrinsic index

xlc (approved):

Issue #564, opened 2021-07-21
orml-vesting claim_for to claim for other accounts


xlc:
Add new extrinsic `claim_for` to allow anyone to trigger claim for any accounts.

PR #568, opened 2021-07-22
Add claim_for

zjb0807:
Closes: #564

### DIFF
--- a/vesting/src/lib.rs
+++ b/vesting/src/lib.rs
@@ -145,13 +145,6 @@ pub mod module {
 		/// The maximum vesting schedules
 		type MaxVestingSchedules: Get<u32>;
 
-		/// A configuration for base priority of unsigned transactions.
-		///
-		/// This is exposed so that it can be tuned for particular runtime, when
-		/// multiple modules send unsigned transactions.
-		#[pallet::constant]
-		type UnsignedPriority: Get<TransactionPriority>;
-
 		// The block number provider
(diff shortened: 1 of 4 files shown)


vesting/src/lib.rs
```diff
@@ -248,6 +255,17 @@ pub mod module {
 			Ok(())
 		}
 
+		/// Verify input by `validate_unsigned
+		#[pallet::weight(T::WeightInfo::claim((<T as Config>::MaxVestingSchedules::get() / 2) as u32))]
+		pub fn claim_for(origin: OriginFor<T>, dest: <T::Lookup as StaticLookup>::Source) -> DispatchResult {
+			ensure_none(origin)?;
```
xlc, line 261: no need to make this unsigned tx.
this is useful to claim rewards for cold wallets that using custom signing software so they don't need to support `vesting.claim`. instead, another account can trigger claim and pay for tx fee.

vesting/src/lib.rs
```diff
@@ -248,6 +255,17 @@ pub mod module {
 			Ok(())
 		}
 
+		/// Verify input by `validate_unsigned
+		#[pallet::weight(T::WeightInfo::claim((<T as Config>::MaxVestingSchedules::get() / 2) as u32))]
+		pub fn claim_for(origin: OriginFor<T>, dest: <T::Lookup as StaticLookup>::Source) -> DispatchResult {
```
xlc, line 260: make this the last transaction to avoid breaking tx index.

xlc (commented):
move thi

xlc (approved):

PR #576, opened 2021-08-04
Update enum serde to use camelCase.

shaunxw:


shaunxw:
structs don't need this.

PR #598, opened 2021-09-01
Limit Call size to 200 bytes

ferrell-code:
The idea of this PR is to limit call length to 200 bytes for all orml pallets. It mirrors https://github.com/paritytech/substrate/pull/9418

Companion to https://github.com/AcalaNetwork/Acala/pull/1386

### DIFF
--- a/authority/src/lib.rs
+++ b/authority/src/lib.rs
@@ -267,7 +267,7 @@ pub mod module {
 		#[pallet::weight(T::WeightInfo::fast_track_scheduled_dispatch())]
 		pub fn fast_track_scheduled_dispatch(
 			origin: OriginFor<T>,
-			initial_origin: T::PalletsOrigin,
+			initial_origin: Box<T::PalletsOrigin>,
 			task_id: ScheduleTaskIndex,
 			when: DispatchTime<T::BlockNumber>,
 		) -> DispatchResult {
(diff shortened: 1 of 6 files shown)


xlc (approved):

ferrell-code:
I could add a integration test that checks every pallet, although there will be a test in Karura/mandala that would catch any large Calls. 

It seems that the XCM stuff has some enums that allocate (relatively) a lot memory as arguments for the extrinsics

authority/src/lib.rs
```diff
@@ -267,7 +267,7 @@ pub mod module {
 		#[pallet::weight(T::WeightInfo::fast_track_scheduled_dispatch())]
 		pub fn fast_track_scheduled_dispatch(
 			origin: OriginFor<T>,
-			initial_origin: T::PalletsOrigin,
+			initial_origin: Box<T::PalletsOrigin>,
```
ferrell-code, line 270: So that decreases the call size from 608 bytes to 24 bytes. Not entirely sure why but numbers don't lie

shaunxw (approved):
LGTM

shaunxw:
Yeah it would be helpful to add integration tests. If there're generic typed params in dispatchable calls, call size would likely to be different between mocks and prod runtimes.

PR #603, opened 2021-09-09
authorize root call

### LINKED ISSUE #602: orml-authority: authorize_root_call
- Storage
  - SavedCalls: map Hash => (Call, Option<AccountId>)
- Call
  - fn authorize_call(origin, call: Call, caller: Option<AccountId>)
    - ensure root
    - save call
  - fn trigger_call(origin, hash: T::Hash)
    - ensure signed
    - take call by hash, and check caller, if saved account is None, allow anyone
    - dispatch call with root origin

This allow us to use governance to schedule something and later dispatched via normal transaction.
Incase something goes wrong, because the call is triggered by a transaction rather hook, it will be possible to purge this bad transaction from mempool to prevent network stall

ermalkaleci:
Closes #602


xlc (commented):
just need a way to cancel saved call and should be good

authority/src/lib.rs
```diff
@@ -325,5 +338,31 @@ pub mod module {
...
+
+		#[pallet::weight(0)]
+		pub fn authorize_call(
+			origin: OriginFor<T>,
+			call: Box<CallOf<T>>,
+			caller: Option<T::AccountId>,
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			let hash = T::Hashing::hash_of(&call);
+			SavedCalls::<T>::insert(hash, (call, caller.clone()));
```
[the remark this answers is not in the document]
ermalkaleci: you can set only same call so it's better do just do one write rather than 1read+1write

xlc (commented):
Need benchmarking & weights and should be good.

authority/src/lib.rs
```diff
@@ -325,5 +339,55 @@ pub mod module {
 			Self::deposit_event(Event::Cancelled(*initial_origin, task_id));
 			Ok(())
 		}
+
+		#[pallet::weight(0)]
```
xlc, line 343: need to update to use weight function

authority/src/lib.rs
```diff
@@ -325,5 +339,55 @@ pub mod module {
...
+
+		#[pallet::weight(T::WeightInfo::authorize_call())]
+		pub fn authorize_call(
+			origin: OriginFor<T>,
+			call: Box<CallOf<T>>,
+			caller: Option<T::AccountId>,
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			let hash = T::Hashing::hash_of(&call);
+			SavedCalls::<T>::insert(hash, (call, caller.clone()));
```
xlc, line 351:
```suggestion
			SavedCalls::<T>::insert(hash, (call, &caller));
```

xlc (approved):

authority/src/lib.rs
```diff
@@ -325,5 +339,55 @@ pub mod module {
...
+
+		#[pallet::weight(T::WeightInfo::trigger_call())]
+		pub fn trigger_call(origin: OriginFor<T>, hash: T::Hash) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			SavedCalls::<T>::try_mutate_exists(hash, |maybe_call| {
+				let (call, maybe_caller) = maybe_call.take().ok_or(Error::<T>::CallNotAuthorized)?;
+				if let Some(caller) = maybe_caller {
+					ensure!(who == caller, Error::<T>::TriggerCallNotPermitted);
+				}
+				let result = call.dispatch(OriginFor::<T>::root());
```
apopiak, line 386: you skip weighing the call

ermalkaleci: > you skip weighing the call
> you'll probably want to pass a weight as param and then ensure that the executed call has lower or equal weight to it

pushed an update to handle call post_info

authority/src/lib.rs
```diff
@@ -325,5 +339,68 @@ pub mod module {
...
+						let caller = maybe_caller.ok_or(Error::<T>::CallNotAuthorized)?;
+						ensure!(who == caller, Error::<T>::CallNotAuthorized);
+					}
+				}
+				Self::deposit_event(Event::RemovedAuthorizedCall(hash));
+				Ok(())
+			})
+		}
+
+		#[pallet::weight(T::WeightInfo::trigger_call())]
```
xlc, line 378: Needs to include the weight of the call. Add new `#[pallet::compact] weight_bound: Weight` and add it to the weight, and then ensure the call weight is lte the weight bound.
Something similar to this
https://github.com/paritytech/substrate/blob/ce1746ddf6588c8f9ecfb77411c3cac7a2c09db3/frame/collective/src/lib.rs#L662

xlc (changes requested):

authority/src/lib.rs
```diff
@@ -375,14 +377,22 @@ pub mod module {
...
+			#[pallet::compact] call_weight_bound: Weight,
+		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 			SavedCalls::<T>::try_mutate_exists(hash, |maybe_call| {
 				let (call, maybe_caller) = maybe_call.take().ok_or(Error::<T>::CallNotAuthorized)?;
 				if let Some(caller) = maybe_caller {
 					ensure!(who == caller, Error::<T>::TriggerCallNotPermitted);
 				}
+				ensure!(
+					call_weight_bound <= call.get_dispatch_info().weight,
```
xlc, line 393: should be >=

xlc (approved):

PR #604, opened 2021-09-10
sweep dust method

ermalkaleci:


tokens/src/lib.rs
```diff
@@ -509,6 +512,23 @@ pub mod module {
...
+
+		#[pallet::weight(0)]
+		pub fn sweep_dust(
+			origin: OriginFor<T>,
+			currency_id: T::CurrencyId,
+			accounts: Vec<T::AccountId>,
+		) -> DispatchResult {
+			T::SweepOrigin::ensure_origin(origin)?;
+			for account in accounts {
+				let balance = Self::free_balance(currency_id, &account);
```
xlc, line 524: should be total balances

ermalkaleci: @xlc this will end up calling `do_transfer` which transfers free balance so if account has frozen or reserved it will remain in storage. Maybe we need to make sure `total_balance` is equal to `free_balance` before calling `on_dust` otherwise account will still exist

xlc: ok. can just require reserved balance is zero.

xlc (commented):
also needs benchmarking & weights and should be good

xlc (approved):

PR #605, opened 2021-09-15
Remove trailing whitespace

ferrell-code:
My editor strips trailing white space and keeps removing this space :)


xlc:
it is a rustfmt thing. you need to either config your editor or config rustfmt

ferrell-code:
yes I see now that it is not a mistake thanks!

PR #609, opened 2021-09-16
make authorize_call free and operational

### LINKED ISSUE #608: Make trigger_call operational and not paying fees

ermalkaleci:
Closes #608

### DIFF
--- a/authority/src/lib.rs
+++ b/authority/src/lib.rs
@@ -20,7 +20,7 @@
 #![allow(clippy::unused_unit)]
 
 use frame_support::{
-	dispatch::PostDispatchInfo,
+	dispatch::{DispatchResultWithPostInfo, PostDispatchInfo},
 	pallet_prelude::*,
 	traits::{
 		schedule::{DispatchTime, Named as ScheduleNamed, Priority},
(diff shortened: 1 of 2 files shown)


authority/src/lib.rs
```diff
@@ -342,7 +342,11 @@ pub mod module {
 			Ok(())
 		}
 
-		#[pallet::weight(T::WeightInfo::authorize_call())]
+		#[pallet::weight((
+			T::WeightInfo::authorize_call(),
+			DispatchClass::Operational,
+			Pays::No,
```
xlc, line 348: I am not 100% sure if this will still charge fee if this call failed. Can you verify?

ermalkaleci: ok

xlc: I know it is safe to do like this https://github.com/open-web3-stack/open-runtime-module-library/blob/69426e8597ebc7c7922b81dbb2f6cc2ab7851e86/oracle/src/lib.rs#L165
An error will not return Pays::No.

ermalkaleci: @xlc yes it will

ermalkaleci: Post dispatch info will be `{ post_info: PostDispatchInfo { actual_weight: None, pays_fee: Pays::Yes }`

xlc (approved):

PR #614, opened 2021-09-20
XCM v1

shaunxw:


xcm-support/src/tests.rs
```diff
@@ -19,12 +19,17 @@ impl Convert<MultiLocation, Option<TestCurrencyId>> for CurrencyIdConvert {
...
-		match l {
-			X1(Parent) => Some(RelayChainToken),
-			X3(Parent, Parachain(1), GeneralKey(k)) if k == token_a => Some(TokenA),
-			X3(Parent, Parachain(2), GeneralKey(k)) if k == token_b => Some(TokenB),
-			_ => None,
+
+		if l == MultiLocation::parent() {
+			return Some(RelayChainToken);
+		}
+		if l == MultiLocation::new(1, X2(Parachain(1), GeneralKey(token_a))) {
```
girazoki, line 26: Why removing the match? And again why not use the former RelativeLocations trait for generating the sybling parachain general keys?

xtokens/src/lib.rs
```diff
@@ -248,92 +251,101 @@ pub mod module {
 			asset: MultiAsset,
 			dest: MultiLocation,
 			recipient: MultiLocation,
-			buy_order: Order<()>,
-		) -> Xcm<T::Call> {
-			WithdrawAsset {
-				assets: vec![asset],
+			dest_weight: u64,
```
girazoki, line 254:
```suggestion
			dest_weight: Weight,
```
I would use Weight in every function

girazoki:
Very nice work @shaunxw . We were in the process of upgrading xtokens to v1 too, but we were wondering how to split the execution fees when we transfer to non-reserve chains. I think the half-half solution that you implemented makes a lot of sense. I just had some minor comments, but other than that looks good

traits/src/location.rs
```diff
@@ -46,77 +47,87 @@ pub trait Reserve {
...
+			Some(MultiLocation::parent())
 		);
 	}
 
 	#[test]
 	fn sibling_parachain_as_reserve_chain() {
 		assert_eq!(
-			concrete_fungible(MultiLocation::X3(Parent, PARACHAIN, GENERAL_INDEX)).reserve(),
-			Some((Parent, PARACHAIN).into())
+			concrete_fungible(MultiLocation::new(1, X2(PARACHAIN, GENERAL_INDEX))).reserve(),
```
girazoki, line 90: You could use the RelativeLocations trait here

xlc: I think this is makes a bit easier to reason the test

xcm-support/src/tests.rs
xlc: I don't think match will work with `MultiLocation::parent()`

xlc (approved):

Issue #616, opened 2021-09-23
tokens dependency conflict


frisitano:
All of `tokens` dependencies are aligned with substrate v4.0.0-dev (branch = "polkadot-v0.9.9") with the exception of [pallet-elections-phragmen](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/tokens/Cargo.toml#L25) which is pointing to a revision associated with substrate 3.0.0. This is causing issues when compiling for substrate v4.0.0-dev.  Can we bump pallet-elections-phragmen to `branch = "polkadot-v0.9.9"` to fix this?

PR #618, opened 2021-09-23
bump pallet-elections-phragmen branch ref for tokens

frisitano:
This PR is to bump the version of `pallet-elections-phragmen` used in `tokens`.  This addresses dependency conflicts as the pre-existing rev reference is associated with substrate v3.0.0. 

Fixes: #616


shaunxw (approved):
Thanks!

PR #619, opened 2021-09-25
fungible helpers

xlc:


tokens/src/impls.rs
```diff
@@ -0,0 +1,209 @@
...
+		if TestKey::contains(&asset) {
+			A::burn_from(dest, amount)
+		} else {
+			B::burn_from(asset, dest, amount)
+		}
+	}
+}
+
+pub trait ConvertBalance<A, B> {
+	fn convert_balance(amount: A) -> B;
```
xlc, line 124: maybe we should just add currency id to here so this can be used to handle multiple rebase tokens.

frankyin-factual: done

PR #644, opened 2021-11-01
Support multiple vesting schedules for the same account at genesis

nazar-pc:
This functionality was already supported in `update_vesting_schedules()`, but not in genesis configuration for some reason

### DIFF
--- a/vesting/src/lib.rs
+++ b/vesting/src/lib.rs
@@ -201,25 +201,30 @@ pub mod module {
 			self.vesting
 				.iter()
 				.for_each(|(who, start, period, period_count, per_period)| {
-					let total = *per_period * Into::<BalanceOf<T>>::into(*period_count);
-
-					let bounded_schedule: BoundedVec<VestingScheduleOf<T>, T::MaxVestingSchedules> =
-						vec![VestingSchedule {
+					let mut bounded_schedules = VestingSchedules::<T>::get(who);
+					bounded_schedules
+						.try_push(VestingSchedule {
 							start: *start,
 							period: *period,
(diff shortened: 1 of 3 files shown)


vesting/src/lib.rs
```diff
@@ -359,17 +364,17 @@ impl<T: Config> Pallet<T> {
...
 	}
+}
 
-	/// Returns `Ok(amount)` if valid schedule, or error.
-	fn ensure_valid_vesting_schedule(schedule: &VestingScheduleOf<T>) -> Result<BalanceOf<T>, DispatchError> {
-		ensure!(!schedule.period.is_zero(), Error::<T>::ZeroVestingPeriod);
-		ensure!(!schedule.period_count.is_zero(), Error::<T>::ZeroVestingPeriodCount);
-		ensure!(schedule.end().is_some(), ArithmeticError::Overflow);
+/// Returns `Ok(total_total)` if valid schedule, or error.
+fn ensure_valid_vesting_schedule<T: Config>(schedule: &VestingScheduleOf<T>) -> Result<BalanceOf<T>, DispatchError> {
```
xlc, line 370: why move this?

xlc:
We don't have this needs to support multiple vesting schedule in genesis but yeah it is good to support this.

vesting/src/lib.rs
nazar-pc: It is now reused in `GenesisConfig` impl, not just in pallet.

nazar-pc:
I have a need for this :wink:

xlc (approved):

PR #645, opened 2021-11-02
avoid reserve and unreserve trigger unexpected Endowed event

### LINKED ISSUE #641: Unexpected Endowed event when reserve and unreserve
`reserve` ：

https://github.com/open-web3-stack/open-runtime-module-library/blob/38b53d39a59c230f4d80d8d8da60cd8c14073523/tokens/src/lib.rs#L1137-L1140

`unreserve`：
https://github.com/open-web3-stack/open-runtime-module-library/blob/38b53d39a59c230f4d80d8d8da60cd8c14073523/tokens/src/lib.rs#L1157-L1158

and maybe there are other functions will write the same `AccountData` more than once, may trigger the `Endowed` event, which is not expected. Consider refactoring the implementation to avoid this situation.

wangjj9219:
close #641

### DIFF
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -1133,14 +1133,12 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
 		}
 		Self::ensure_can_withdraw(currency_id, who, value)?;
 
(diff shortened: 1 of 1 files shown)


xlc:
we need to have some tests

xlc (approved):

PR #651, opened 2021-11-10
relay account and tests

### LINKED ISSUE #636: RelayChainAccountId32Aliases
Add `RelayChainAccountId32Aliases` to xcm-support.

Similar to `AccountId32Aliases` from Polkadot, but also accept AccountId32 from relaychain.

Need to make sure not accepting AccountId32 from other parachains.

Need integration tests.

zqhxuyuan:
- add `RelaychainAccountId32Aliases` to xcm-support
- add `AllowRelayedPaidExecutionFromParent` to xcm-support
- add unit testcase
- add integration test in acala: https://github.com/AcalaNetwork/Acala/pull/1614

`AllowRelayedPaidExecutionFromParent` is needed because in the case of relaychain account send `Transact` to parachain, then `DescendOrigin` is automatically added as the first instruction sended to parachain. so it's used as Barrier check to allow this situation passing through.

`RelaychainAccountId32Aliases` is used as convert relaychain account to account controlled on parachain. and it's related to above case when `DescendOrigin` append `origin`(which is `Parent`) with `Account` which result to `(Parent, Account)`.

closes https://github.com/open-web3-stack/open-runtime-module-library/issues/636


shaunxw (changes requested):
Any reason to add this struct in `orml-xcm-support`? It looks like a runtime config to me.

xlc:
it is something to be used by all the runtime and potentially used by other parachains.
Like https://github.com/paritytech/polkadot/blob/54d1cb9178fd41b535d8d908b021926d01d91f00/xcm/xcm-builder/src/location_conversion.rs#L104

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,75 @@ impl UnknownAsset for () {
...
+impl<Network: Get<NetworkId>, AccountId: From<[u8; 32]> + Into<[u8; 32]> + Clone>
+	xcm_executor::traits::Convert<MultiLocation, AccountId> for RelaychainAccountId32Aliases<Network, AccountId>
+{
+	fn convert(location: MultiLocation) -> Result<AccountId, MultiLocation> {
+		let id = match location {
+			MultiLocation {
+				parents: 1,
+				interior: X1(AccountId32 {
+					id,
+					network: NetworkId::Any,
```
KiChjang, line 94: This means any account on the parent network is going to be able to pass through this filter. The assumption here is that the parent network is always the relay chain, which doesn't hold true in all contexts. Instead, I propose adding another type parameter called `RelayChainLocation` to specifically define where exactly the relay chain is. It should be trait bounded with a `Contains`.

KiChjang: This is also missing a match arm where the network ID here is equal to `Network::get()`, instead of being `Network::Any`.

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,75 @@ impl UnknownAsset for () {
...
+		origin: &MultiLocation,
+		message: &mut Xcm<Call>,
+		max_weight: Weight,
+		_weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		ensure!(T::contains(origin), ());
+		let mut iter = message.0.iter_mut();
+		let i = iter.next().ok_or(())?;
+		match i {
+			DescendOrigin(..) => (),
```
KiChjang, line 123: I'm struggling to see what the purpose of this is. Is it trying to allow for lowering privileges? If so, then this should check whether or not the `InteriorMultiLocation` inside of the `DescendOrigin` is `Here` or not, because a `Here` would not lower any privileges.

Would also help if there are docs detailing what `AllowDescendOrigin` seeks to achieve.

zqhxuyuan: in one of our user case(the testcase is [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/relay_account/xtokens/src/tests.rs#L460)), the xcm format is: `DescendOrigin | WithdrawAsset | BuyExecution | Transact`, if the Barrier is `AllowTopLevelPaidExecutionFrom`, then it would't passing Barrier check. so we need `AllowDescendOrigin`.

I haven't consider about the interior location inside [`AllowDescendOrigin`](https://github.com/paritytech/polkadot/blob/release-v0.9.12/xcm/xcm-executor/src/lib.rs#L335-L340), in my testcase, the `who` of  `AllowDescendOrigin(who)` is `Alice` account, and the `self.origin` is `Parent`, the result `origin` is `(Parent, Alice)`, that's why we add `RelaychainAccountId32Aliases` also.

xcm-support/src/lib.rs
zqhxuyuan: sounds reasonable

xcm-support/src/lib.rs
KiChjang: But why is there a `DescendOrigin`? What is it trying to achieve here? To check that the origin of the XCM ultimately came from an account that the recipient controls? If so, then the name of this struct needs to reflect that, i.e. you're not simply allowing descend origin, you're allowing accounts owned by the same user to send XCMs to themselves. That sounds like `AllowEquivalentAccountsFrom`.

Moreover, there will also need to be a logic to check whether the recipient account matches the `AccountId` of the `origin` as well.

zqhxuyuan: sorry not detail the background. the xcm message cames like that: a RelayChain acount send `Transact` to a dest Parachain. As the `interior` is not `Here`, but an `Account`, so it'll add `DescendOrigin` in the sender side. that's why dest Parachain got `DescendOrigin` at first. 

> What is it trying to achieve here? To check that the origin of the XCM ultimately came from an account that the recipient controls?

yeap, so much like. in my background written above, when Parachain got the resulting origin as `(Parent, Account)`, we want the origin become to `Account` again when execute `Transact`, that's why `RelaychainAccountId32Aliases` is used, it'll make `(Parent, Account)` transform to `Account`, after that we'll use this parachain Account do our own logic. and here the final resulting `Account` is what you called _an account that the recipient controls_

to sum up, the `AllowDescendOrigin` is for barrier check, the `RelaychainAccountId32Aliases` is used for transform a parent account(sender) to account that parachain(receiver) can control.

the name `AllowDescendOrigin` is unthinkingly, you sugesst of `AllowEquivalentAccountsFrom` sounds good.
we'll rename it.

ps: I'm also thinking/consider about, we may just allow `DescendOrigin | BuyExecution | Transact` instead of `DescendOrigin | WithdrawAsset | BuyExecution | ...`, as our case may need `WithdrawAsset` after `DescendOrigin` and before `Transact`. what do you think? @xlc

xlc: It is best to whitelist as minimal as possible. So it is best if you can reduce allowed actions as much as possible while keeping the whole e2e flow working

KiChjang: I second Bryan's opinion. Casting too wide a net easily leads to a privilege escalation attack.

zqhxuyuan: unfortunately, `DescendOrigin | BuyExecution(limited weight) | Transact` can't work, unless `BuyExecution` use `Unlimited`, then It's just like `DescendOrigin | Transact`. because `BuyExecution(limited)` needs `holding` has assets, orelse it'll cause `NotHoldingFees`. I'll try `DescendOrigin | WithdrawAsset | BuyExecution(limited weight) | Transact`, and only use assets in `WithdrawAsset` for payment fees.

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,79 @@ impl UnknownAsset for () {
...
+		};
+		Ok(id.into())
+	}
+
+	fn reverse(who: AccountId) -> Result<MultiLocation, AccountId> {
+		Ok(AccountId32 {
+			id: who.into(),
+			network: Network::get(),
+		}
+		.into())
```
apopiak, line 107: This will reverse to `MultiLocation {
				parents: 0,
				interior: X1(AccountId32),
}`, no?
That means it will not convert back to the input if calling `RelaychainAccountId32Aliases.convert(location).reverse()` which would be desirable, no?

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,79 @@ impl UnknownAsset for () {
...
+pub struct AllowEquivalentAccountsFrom<T>(PhantomData<T>);
+impl<T: Contains<MultiLocation>> ShouldExecute for AllowEquivalentAccountsFrom<T> {
+	fn should_execute<Call>(
+		origin: &MultiLocation,
+		message: &mut Xcm<Call>,
+		max_weight: Weight,
+		_weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		ensure!(T::contains(origin), ());
+		let mut iter = message.0.iter_mut();
```
apopiak, line 125: Why does it have to be mutable?

KiChjang: Because `Iterator::next` is a mutable method, and it's used in the line below.

xcm-support/src/lib.rs
zqhxuyuan: It's a good point, as the reverse path is not currently go through, I may miss here.

zqhxuyuan: after I change here to return `(1, AccountId32).into()`. If we config `(...,AccountId32Aliases,RelaychainAccountId32Aliases)`, then reverse on AccountId will always return `(parents:0, X1(AccountId32))`. But when I change the order to `(...,RelaychainAccountId32Aliases, AccountId32Aliases)`, then recerse on AccountId will always return `(parents:1, X1(AccountId32))`. this sounds no better solution.

xcm-support/src/lib.rs
```diff
@@ -89,22 +89,22 @@ impl<Network: Get<NetworkId>, AccountId: From<[u8; 32]> + Into<[u8; 32]> + Clone
 		let id = match location {
 			MultiLocation {
 				parents: 1,
-				interior: X1(AccountId32 {
-					id,
-					network: NetworkId::Any,
-				}),
-			} => id,
+				interior: X1(AccountId32 { id, network }),
+			} if network == Network::get() => id,
```
KiChjang, line 93: You needed to _add_ a match arm for the case where `network` is equal to `Network::get()`, not _modify_ the existing one where `network` is set to `Any`. Right now it's missing exactly that case, see https://github.com/paritytech/polkadot/blob/903c6f2f014bcedb9d6be0cae86e6a4a922f692b/xcm/xcm-builder/src/location_conversion.rs#L109-L116 for more information.

xcm-support/src/lib.rs
apopiak: I meant the iterator (`iter_mut`). And I just found the mutation below `*weight = max_weight;` so never mind.

xcm-support/src/lib.rs
zqhxuyuan: @KiChjang do you think we need to add `DescendOrigin(..)` restricted condition here such as `DescendOrigin(X1(Junction::AccountId32(..))`?

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,86 @@ impl UnknownAsset for () {
...
+			.into())
+	}
+}
+
+/// when Relay an XCM message from a given `interior` location, if the given
+/// `interior` is not `Here`, the destination will receive a xcm message
+/// beginning with `DescendOrigin` as the first instruction. so the xcm message
+/// format must match this order:
+/// `DescendOrigin`,`WithdrawAsset`,`BuyExecution`,`Transact`.
+pub struct AllowEquivalentAccountsFrom<T>(PhantomData<T>);
```
xlc, line 123: The same is still a bit confusing. `AllowEquivalentAccountsFrom<Everything> ` is not actually allow from everything.

Maybe `AllowEquivalentParentAccountsFrom`

xtokens/Cargo.toml
```diff
@@ -49,6 +49,8 @@ xcm-simulator = { git = "https://github.com/paritytech/polkadot", branch = "rele
 orml-tokens = { path = "../tokens" }
 orml-xcm = { path = "../xcm" }
 
+env_logger = "0.8.3"
```
shaunxw, line 52: Not needed.

xcm-support/src/lib.rs
zqhxuyuan: ok. I was thinking add restricted condition to DescendOrigin, only allow `DescendOrigin(X1(Junction::AccountId32 { network} if network == Network.get()` and Network trait bound is `RelayNetwork`

shaunxw: The name `AllowRelayedPaidExecutionFrom` might be more expressive.

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,86 @@ impl UnknownAsset for () {
...
+		)
+			.into())
+	}
+}
+
+/// when Relay an XCM message from a given `interior` location, if the given
+/// `interior` is not `Here`, the destination will receive a xcm message
+/// beginning with `DescendOrigin` as the first instruction. so the xcm message
+/// format must match this order:
+/// `DescendOrigin`,`WithdrawAsset`,`BuyExecution`,`Transact`.
```
shaunxw, line 129: The docstring needs update to address the use case. Refer to [`AllowTopLevelPaidExecutionFrom`](https://github.com/paritytech/polkadot/blob/f26005b999acb8936bda66df10daa0d75c85ce62/xcm/xcm-builder/src/barriers.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L48-L52)

xcm-support/src/lib.rs
zqhxuyuan: I also change `Everything` to a new struct `IsParent` to only allow `Relaychain` is the `origin`.

xtokens/src/mock/tests.rs
```diff
@@ -0,0 +1,359 @@
+#![cfg(test)]
```
shaunxw, line 1: Tests in this file look like for testing upstream deps, I don't see why we need them in `orml`.

xtokens/src/tests.rs
```diff
@@ -431,6 +404,309 @@ fn send_as_sovereign_fails_if_bad_origin() {
...
+
+	ParaA::execute_with(|| {
+		parachain_transact_to_relaychian_remark();
+	});
+
+	Relay::execute_with(|| {
+		use relay::{Event, System};
+		assert!(System::events()
+			.iter()
+			.any(|r| matches!(r.event, Event::System(frame_system::Event::Remarked(_, _)))));
```
shaunxw, line 421: Use `System::assert_has_event`, same to all other remark event checks.

xtokens/src/tests.rs
```diff
@@ -431,6 +404,309 @@ fn send_as_sovereign_fails_if_bad_origin() {
...
+	});
+}
+
+#[test]
+fn relay_transact_to_para_remark_use_normal_account() {
+	ParaA::execute_with(|| {
+		assert_ok!(ParaTokens::deposit(CurrencyId::R, &ALICE, 6040));
+		assert_eq!(7040, ParaTokens::free_balance(CurrencyId::R, &ALICE));
```
shaunxw, line 445: Not needed. We don't test for `deposit` fn here. Same to a lot of following balance equality checks.

xcm-support/src/lib.rs
zqhxuyuan: I think we may not need `T: Contains<MultiLocation>` trait bound, we can use more strict condition inside like 
```rust
ensure!(origin.contains_parents_only(1));
```

xtokens/src/mock/tests.rs
zqhxuyuan: move tests here to xcm-support

xtokens/src/tests.rs
zqhxuyuan: delete integration test as the testcase is not scope of xTokens pallet. it will be add to acala repo.

xtokens/src/tests.rs
zqhxuyuan: delete useless tests.

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,84 @@ impl UnknownAsset for () {
...
+		)
+			.into())
+	}
+}
+
+/// Allows execution from `origin` if it is `Parent`.
+///
+/// Only allows exactly alike the order of xcm: `DescendOrigin`,
+/// `WithdrawAsset`, `BuyExecution` and `Transact`.
+pub struct AllowRelayedPaidExecutionFrom<Network>(PhantomData<Network>);
```
KiChjang, line 117: Final nitpick: since you've restricted to parents only, you can just name it `AllowRelayedPaidExecutionFromParent`.

xcm-support/src/lib.rs
zqhxuyuan: I've post an issue on polkadot: https://github.com/paritytech/polkadot/issues/4296. would you mind to see if that's a potential problem? @KiChjang @apopiak

zqhxuyuan:
current the xcm format is `WithdrawAsset | BuyExecution(limited) | Transact`, this means the asset of relaychain acount(say 1 DOT) will all be first move to `holding`, then payment fee will be buyed for execution(say 0.1 DOT), after that unspent asset(0.9 DOT) will be in holding again. and As the xcm has its AssetTrap mechanism, the unspent asset will be lefted in the storage for later claim. Do we need add another `DepositAsset` to do this for users? @xlc

xcm-support/src/tests.rs
```diff
@@ -98,3 +100,50 @@ fn multi_native_asset() {
...
+		},
+		Transact {
+			origin_type: OriginKind::SovereignAccount,
+			require_weight_at_most: 1000 as u64,
+			call: Encode::encode(&100).into(),
+		},
+	]);
+	let r =
+		AllowRelayedPaidExecutionFromParent::<RelayNetwork>::should_execute(&(Parent.into()), &mut xcm, 100, &mut 100);
+	assert_eq!(r, Ok(()));
```
xlc, line 148:
```suggestion
	assert_ok!(r);
```

xlc:
Yeah we should allow `DepositAsset` the unused fees back.

Also maybe should support `ReserveAssetDeposited` so it is possible to send some funds from relay chain for the tx fee. Otherwise user needs to fund the account on parachain first.

zqhxuyuan:
> Yeah we should allow `DepositAsset` the unused fees back.
> 
> Also maybe should support `ReserveAssetDeposited` so it is possible to send some funds from relay chain for the tx fee. Otherwise user needs to fund the account on parachain first.

it's easy add `ReserveAssetDeposited` to match case inside `AllowRelayedPaidExecutionFromParent`, but construct a xcm in relaychain side currently didn't work. e.g. relaychain has xcm format `TransferReserverAsset{xcm: BuyExecution | Transact | DepositAsset }`, then parachain receives: `ReserveAssetDeposited | ClearOrigin | BuyExecution | Transact | DepositAsset`. As the `Transact` instruction need `origin` not null, but previous `ClearOrigin` set `origin` to `None`, this will cause `BadOrigin` error in parachain.

xcm-support/src/lib.rs
```diff
@@ -75,3 +78,82 @@ impl UnknownAsset for () {
...
+/// Allows execution from `origin` if it is `Parent`.
+pub struct AllowRelayedPaidExecutionFromParent<Network>(PhantomData<Network>);
+impl<Network: Get<NetworkId>> ShouldExecute for AllowRelayedPaidExecutionFromParent<Network> {
+	fn should_execute<Call>(
+		origin: &MultiLocation,
+		message: &mut Xcm<Call>,
+		max_weight: Weight,
+		_weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		ensure!(origin.contains_parents_only(1), ());
```
ghzlatarev, line 122: Is it possible to make this a configurable generic that implements `Contains<MultiLocation>` ?
We have a use-case for allowing sibling chains to send `Transact` instructions to us.

PR #658, opened 2021-11-22
Girazoki controlled fee specification

### LINKED ISSUE #655: Be able to specify the fee aside the amount to be transferred in Xtokens
This is a usecase we encountered in Moonbeam, where we need to specify the fee (the amount to be provided in BuyExecution) as a separate argument to the amount being transferred, with the goal of better controlling the amount received in the destination side. We are in the process of adding this to Xtokens in our own fork, would you be interested on having such feature too? I could do a  PR with my changes if you are interested

girazoki:
Adds support for specifying the fee as a separate argument to the amount being transferred in xtokens. The fee amount will be used to pay for the buyExecution order in the destination chain. This is useful for cases in which users (or smart contracts) want to have a good control of the amount of tokens in a particular address. The existing extrinsics substract this fee from the amount being transferred, and thus its difficult to know how much tokens a given address has in a specific chain.

For the aforementioned motivation, non-spent fee amounts will get trapped in the destination chain with the new transfer_with_fee and transfer_multiasset_with_fee extrinsics.

For now I made it such that the fee needs to have the same asset id as the token you are transferring. I thought it might be good to test it like this, while we can relax this assumption in the future

Closes #655


xtokens/src/lib.rs
```diff
@@ -254,6 +345,65 @@ pub mod module {
...
+			fee: MultiAsset,
+			dest: MultiLocation,
+			dest_weight: Weight,
+			deposit_event: bool,
+		) -> DispatchResult {
+			if !asset.is_fungible(None) || !fee.is_fungible(None) {
+				return Err(Error::<T>::NotFungible.into());
+			}
+
+			if fungible_amount(&asset).is_zero() || fungible_amount(&fee).is_zero() {
```
xlc, line 360: zero fee should be an error?

xtokens/src/lib.rs
```diff
@@ -170,6 +176,38 @@ pub mod module {
...
+		/// Transfer native currencies specifying the fee and amount as
+		/// separate.
+		///
+		/// `dest_weight` is the weight for XCM execution on the dest chain, and
+		/// it would be charged from the transferred assets. If set below
+		/// requirements, the execution may fail and assets wouldn't be
+		/// received.
+		///
+		/// `fee` is the amount to be spent to pay for execution in destination
+		/// chain Both fee and amount will be substracted form the callers
```
xlc, line 188:
```suggestion
		/// chain. Both fee and amount will be substracted form the callers
```

xtokens/src/lib.rs
```diff
@@ -170,6 +176,38 @@ pub mod module {
...
+		/// balance
```
xlc, line 189:
```suggestion
		/// balance.
```

xtokens/src/lib.rs
```diff
@@ -104,8 +104,12 @@ pub mod module {
 	pub enum Event<T: Config> {
 		/// Transferred. \[sender, currency_id, amount, dest\]
 		Transferred(T::AccountId, T::CurrencyId, T::Balance, MultiLocation),
+		/// Transferred with fee. \[sender, currency_id, amount, fee, dest\]
+		TransferredWithFee(T::AccountId, T::CurrencyId, T::Balance, T::Balance, MultiLocation),
 		/// Transferred `MultiAsset`. \[sender, asset, dest\]
 		TransferredMultiAsset(T::AccountId, MultiAsset, MultiLocation),
+		/// Transferred `MultiAsset` with fee. \[sender, asset, dest\]
```
xlc, line 111:
```suggestion
		/// Transferred `MultiAsset` with fee. \[sender, asset, fee, dest\]
```

xtokens/src/lib.rs
```diff
@@ -170,6 +176,38 @@ pub mod module {
...
+		/// balance
+		///
+		/// It's a no-op if any error on local XCM execution or message sending.
+		/// Note sending assets out per se doesn't guarantee they would be
+		/// received. Receiving depends on if the XCM message could be delivered
+		/// by the network, and if the receiving chain would handle
+		/// messages correctly.
+		#[pallet::weight(Pallet::<T>::weight_of_transfer(currency_id.clone(), *amount, dest))]
+		#[transactional]
+		pub fn transfer_with_fee(
```
xlc, line 198: all the new calls must be added after existing ones otherwise it is a transaction version breaking change.

xlc:
Interesting that you are willing to burn the extra fee rather than give it back to user but I guess I don't have a strong reason to stop you from doing that.

xtokens/src/lib.rs
```diff
@@ -134,6 +138,8 @@ pub mod module {
 		/// The version of the `Versioned` value used is not able to be
 		/// interpreted.
 		BadVersion,
+		/// The fee MultiAsset is of different type than the asset to transfer
```
shaunxw, line 141:
```suggestion
		/// The fee MultiAsset is of different type than the asset to transfer.
```

xtokens/src/lib.rs
```diff
@@ -195,6 +233,40 @@ pub mod module {
...
+		///
+		/// `dest_weight` is the weight for XCM execution on the dest chain, and
+		/// it would be charged from the transferred assets. If set below
+		/// requirements, the execution may fail and assets wouldn't be
+		/// received.
+		///
+		/// `fee` is the multiasset to be spent to pay for execution in
+		/// destination chain Both fee and amount will be substracted form the
+		/// callers balance For now we only accept fee and asset having the same
+		/// multilocation id
```
shaunxw, line 247:
```suggestion
		/// destination chain. Both fee and amount will be subtracted form the
		/// callers balance For now we only accept fee and asset having the same
		/// `MultiLocation` id.
```

xtokens/src/lib.rs
shaunxw:
```suggestion
		/// chain Both fee and amount will be subtracted form the callers
```

shaunxw (commented):
It's worth noting that `fee` would be burned instead of returned to users in docstring. I'm curious about the reason to burn them.

xtokens/src/lib.rs
girazoki: perfect

girazoki:
The reason for the fee to be burned (or trapped in the destination chain, whatever the destination chain decides to do with non-spent money in the holding register) is that the use case we are trying to cover is that in which:

- A user (or smart contract) wants to have control over how much tokens the destination account has.
- The fee that the operation will spend might not be known in advance and therefore the user is willing to  slightly overpay and burn the non-spent fee amount

girazoki:
That is actually very right. In that case then maybe we can return the non-used to the deposit account, and what the caller would have is a sense of "minimum balance" that the deposited account should have.

Will modify the PR accordingly. Thanks @xlc!

xtokens/src/lib.rs
girazoki: Initially I thought this could be good, but it might prevent some usecases where you transfer to free-execution chains. I will remove it

xlc (commented):
I think it should error when fee is zero and test case for that.
And some comment indicate if fee is too low, then the asset will be trapped, and some test to check the asset is trapped if that's not too hard.

girazoki:
yeah no problem, I can do that

xlc (approved):

xtokens/src/lib.rs
xlc: I guess we can revisit this later if there is indeed a chain charges zero fee for buying weights.

girazoki:
Is there any reason the tests fail? they work locally for me

girazoki:
I did shortcut also in do_transfer_with_fee for fee==0, before converting it into a fungible, since I think that was the origin of the tests panicking

shaunxw (approved):

PR #659, opened 2021-11-23
xtokens: ensuring exact error from xcm propagated

dzmitry-lahoda:
xcmp may fail because of dozens of reasons. it is very hard to debug and hard to setup. 
so solution is to be very specific about failure reason.

### DIFF
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -134,6 +134,91 @@ pub mod module {
 		/// The version of the `Versioned` value used is not able to be
 		/// interpreted.
 		BadVersion,
+		/// same as  XCM underlying error
+		XcmOverflow,
+		/// same as  XCM underlying error
+		XcmUnimplemented,
+		/// same as  XCM underlying error
+		XcmUntrustedReserveLocation,
+		/// same as  XCM underlying error
+		XcmUntrustedTeleportLocation,
+		/// same as  XCM underlying error
+		XcmMultiLocationFull,
+		/// same as  XCM underlying error
+		XcmMultiLocationNotInvertible,
+		/// same as  XCM underlying error
+		XcmBadOrigin,
+		/// same as  XCM underlying error
+		XcmInvalidLocation,
+		/// same as  XCM underlying error
+		XcmAssetNotFound,
+		/// same as  XCM underlying error
+		XcmFailedToTransactAsset,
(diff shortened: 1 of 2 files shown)


xtokens/src/lib.rs
```diff
@@ -245,7 +330,7 @@ pub mod module {
 			let weight = T::Weigher::weight(&mut msg).map_err(|()| Error::<T>::UnweighableMessage)?;
 			T::XcmExecutor::execute_xcm_in_credit(origin_location, msg, weight, weight)
 				.ensure_complete()
-				.map_err(|_| Error::<T>::XcmExecutionFailed)?;
+				.map_err(xcm_to_pallet_error::<T>)?;
```
zjb0807, line 333: I think it is enough to print the log here.

dzmitry-lahoda: may be. 
may be you know how to get logs in xcm simulator tests https://stackoverflow.com/questions/70085232/how-to-see-logs-out-of-tests-running-substrate-cumulus-based-runtime ?

xlc (approved):
I am fine with this. Later can be updated once https://github.com/paritytech/substrate/pull/10242 is available

xtokens/src/lib.rs
zjb0807: 1. add `env_logger::init();` in the top of test method.
2. Run with
```
RUST_LOG=trace,parity-db=error,trie=error cargo test
```

dzmitry-lahoda:
replaced with logger. added comment to use paritytech/substrate#10242  as soon as ready.

xtokens/src/lib.rs
```diff
@@ -245,7 +247,10 @@ pub mod module {
 			let weight = T::Weigher::weight(&mut msg).map_err(|()| Error::<T>::UnweighableMessage)?;
 			T::XcmExecutor::execute_xcm_in_credit(origin_location, msg, weight, weight)
 				.ensure_complete()
-				.map_err(|_| Error::<T>::XcmExecutionFailed)?;
+				.map_err(|error| {
+					log::trace!("Failed execute transfer message with {:?}", error);
```
zjb0807, line 251:
```suggestion
					log::error!(target: "xtokens", "Failed execute transfer message with {:?}", error);
```

PR #677, opened 2022-01-10
Updated test coverage script

syan095:
Updated action versions
Changed to use self-hosted


xlc:
Can we only run it for GH provided instances? We only have limited number of self-hosted instances and I don't want tasks to be queued for too long.

zjb0807 (approved):

PR #678, opened 2022-01-14
Allow to check arbitrary PreConditions before transfers

mustermeiszer:
This is a proposal to allow runtimes to run arbitrary checks on a given `sender, receiver, currency_id, amount` combination for normal transfers. 

The current approach only checks if the preconditions for a transfer are met when using the public API. I.e. only for normal users. 
The trait interfaces of Multicurrency, Fungibles, etc are not checked. I am not sure if this is generally wanted or not. Substrate tends to leave traits pretty much permissionless.

Happy for feedback and a temperature check how the acceptance of this is.

### DIFF
--- a/traits/src/lib.rs
+++ b/traits/src/lib.rs
@@ -91,7 +91,7 @@ pub trait PreConditions<T> {
 	fn check(t: &T) -> bool;
 }
 
-#[impl_for_tuples(10)]
+#[impl_for_tuples(1, 10)]
 impl<T> PreConditions<T> for Tuple {
 	fn check(t: &T) -> bool {
 		for_tuples!( #( Tuple::check(t) )&* )
--- a/traits/src/lib.rs
+++ b/traits/src/lib.rs
@@ -100,16 +100,16 @@ impl<T> PreConditions<T> for Tuple {
 
 #[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, TypeInfo)]
 pub struct Always;
-impl PreConditions<T> for Always {
-	fn check(t: &T) -> bool {
+impl<T> PreConditions<T> for Always {
+	fn check(_t: &T) -> bool {
 		true
 	}
 }
 
 #[derive(Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug, TypeInfo)]
 pub struct Never;
-impl PreConditions<T> for Never {
-	fn check(t: &T) -> bool {
+impl<T> PreConditions<T> for Never {
+	fn check(_t: &T) -> bool {
 		false
 	}
 }
--- a/traits/src/lib.rs
+++ b/traits/src/lib.rs
@@ -86,3 +86,24 @@ impl<T> Handler<T> for Tuple {
 		Ok(())
 	}
 }
+
+#[impl_for_tuples(30)]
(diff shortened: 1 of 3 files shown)


xlc (commented):
Another way to achieve similar functionality without modification here is simply do not expose the calls from orml-tokens, and write your own pallet that exposes calls that does its custom check and call transfer on orml-tokens. Something a bit similar to orml-currencies.

traits/src/lib.rs
```diff
@@ -86,3 +86,30 @@ impl<T> Handler<T> for Tuple {
 		Ok(())
 	}
 }
+
+pub trait PreConditions<T> {
```
xlc, line 90: this is basically the `Filter` trait from Substrate

xlc:
I will say you really should do the permission control on exposed calls on your own pallet. If you want to enforce some precondition on the fungible traits level, then I will say implement some type helper similar to `Combiner` or `Mapper` like https://github.com/open-web3-stack/open-runtime-module-library/blob/master/tokens/src/impls.rs

mustermeiszer:
> Another way to achieve similar functionality without modification here is simply do not expose the calls from orml-tokens, and write your own pallet that exposes calls that does its custom check and call transfer on orml-tokens. Something a bit similar to orml-currencies.

Yes, you are absolutely right. This is what we currently are doing. We basically mimic the Call-interface of tokens and do our pre-checks there and then forward to fungibles implementations.

 I have one main reasoning for bringing up this PR here.
* Client compatibility/integrations

WIth the two-pallets approach, clients (wallets, etc.) must always query the storage of orml-tokens for balances and know the Call-Api of the "guarding" pallet for transfers. Even if we mimic the orml-tokens interface 1-to-1 and clients must know the location of orml-tokens. Doing this manually for every chain is an unnecessary overhead. 

With the proposed change, the orml-tokens could be used as the default tokens pallet for almost every use-case and client integrations would simply switch the ws-endpoint from chain to chain (assuming everybody defaults to naming the orml-tokens pallet `Tokens`).

traits/src/lib.rs
mustermeiszer: Yeah, but parity seems to abandon this one: https://github.com/paritytech/substrate/blob/3168e10f6576a28930554bbf903ff9900b833a14/frame/support/src/traits/members.rs#L44-L51

xlc:
we are not going to get everyone to use this pallet. My solution to unified interface is https://github.com/w3f/PSPs/blob/master/PSPs/drafts/psp-3.md

mustermeiszer:
But when will this be realized? I still think this rather small change to orml-tokens allows for broader adoption, which at the beginning will make integrations easier.

xlc:
It is really not a good idea to use calls of orml-tokens directly anyway. Most of the teams will want to use orml-currencies instead so that a unified interface can be used to transfer native token and custom token. From my 1 min research, Karura, Bifrost, Basilisk are all using orml-tokens but expose currencies for transaction.

mustermeiszer:
I see your point. But at least HydraDx and Bifrost also expose orml-tokens. I have to think about it.

mustermeiszer:
I am closing this. @xlc I will wrap ours in a restricted-currency version. Thanks for the input and advice!

PR #683, opened 2022-01-25
Girazoki dual token transferrable

girazoki:
This is a draft proposal (still many things to polish) for being able to send distinct asset and fee assets, as long as they have the same reserve chain. This is intended for chains that only accept a small subset of assets as for xcm fee payment. For instance. imagine Parachain A has `pallet_assets` with an asset `AssetA` and `pallet_balances` with its native token is` NativeA`.

A reciving parachain B would only need to support xcm payments in` NativeA`, and thus if one wants to send `AssetA` it would need to send also `NativeA` for fee payment. This is intended to potentially limit the assets a parachain supports for fee payment.

Additionally, a refactor is included in preparation for the possibility to, in the future, send more than just 2 tokens to a chain


xtokens/src/lib.rs
```diff
@@ -254,18 +256,19 @@ pub mod module {
 			origin: OriginFor<T>,
 			currency_id: T::CurrencyId,
 			amount: T::Balance,
-			fee: T::Balance,
+			fee_id: T::CurrencyId,
```
xlc, line 259: can you add new method instead modify existing one, otherwise it will be a major transaction version breaking change.

girazoki: absolutely

xtokens/src/lib.rs
```diff
@@ -421,34 +504,90 @@ pub mod module {
...
+			let mut assets = MultiAssets::new();
+			for (currency_id, amount) in currencies.clone() {
+				let location: MultiLocation = T::CurrencyIdConvert::convert(currency_id.clone())
+					.ok_or(Error::<T>::NotCrossChainTransferableCurrency)?;
+				// Push contains saturated addition, so we should be able to use it safely
+				assets.push((location.clone(), amount.into()).into())
+			}
+
+			// We first grab the fee
+			let fee = assets.get(fee_item as usize).ok_or(Error::<T>::Empty)?;
```
girazoki, line 533: Maybe should be taken before we merge?

girazoki:
Several tests to be added yet (specially for errors) but major changes have been:

- A unique function where we calculate the transfer_kind, and also unique instruction functions to which we pass multiassets now.
- Addition of two new extrinsics that allow to transfer up to two distinct assets as long as they share the same reserve
- Removal of the ArithmeticOverflow error and use the MultiAssets push function instead, which already uses saturated addition

xtokens/src/lib.rs
```diff
@@ -100,6 +100,10 @@ pub mod module {
 		type LocationInverter: InvertLocation;
 	}
 
+	/// The maximum number of distinct assets allowed to be transferred in a
+	/// single helper extrinsic.
+	const MAX_ASSETS_FOR_TRANSFER: usize = 2;
```
shaunxw, line 105: I think this could be a type in `Config`, to be more flexible.

xtokens/src/lib.rs
```diff
@@ -160,11 +176,16 @@ pub mod module {
...
+		/// We tried sending distinct asset and fee but they have different
+		/// reserve chains
+		DistinctReserveForAssetAndFee,
 		/// The fee amount was zero when the fee specification extrinsic is
 		/// being used.
 		FeeCannotBeZero,
+		/// The number of assets to be sent is over the maximum
+		TooManyAssets,
+		/// The specified item does not exist in the vec assets
+		Empty,
```
shaunxw, line 188: Can we have more explicit naming

shaunxw (approved):

xlc (approved):

PR #684, opened 2022-01-27
Feature/nft-trait

syan095:
Module NFT implemented new traits to mint and burn NFT.
Added a new function in the NFT trait


traits/src/nft.rs
```diff
@@ -27,3 +27,26 @@ pub trait NFT<AccountId> {
 	/// Transfer the given token ID from one account to another.
 	fn transfer(from: &AccountId, to: &AccountId, token: (Self::ClassId, Self::TokenId)) -> DispatchResult;
 }
+
+// This trait provides interface to manage NFTs
+#[allow(clippy::upper_case_acronyms)]
+pub trait ManageNFT<AccountId, CID, Attributes> {
```
xlc, line 33: make it `NftMinter: NFT` and `NftBurner: NFT` so no need to duplicate `type ClassId`

zqhxuyuan:
do we need let module `nft` implements this new trait?

traits/src/nft.rs
syan095: Split this into seperate traits

traits/src/nft.rs
```diff
@@ -20,10 +17,6 @@ pub trait NFT<AccountId> {
 	/// The number of NFTs assigned to `who`.
 	fn balance(who: &AccountId) -> Self::Balance;
 
-	/// The owner of the given token ID. Returns `None` if the token does not
-	/// exist.
-	fn owner(token: (Self::ClassId, Self::TokenId)) -> Option<AccountId>;
```
xlc, line 25: why remove those?

syan095: The Inspect trait contains a "owner" function and the "transfer" trait contains "transfer". If we keep both, it introduces extra confusion. So I thought it would be more "clean" to remove them so there's only one transfer() and one owner()

traits/src/lib.rs
```diff
@@ -19,7 +19,7 @@ pub use currency::{
 pub use data_provider::{DataFeeder, DataProvider, DataProviderExtended};
 pub use get_by_key::GetByKey;
 pub use multi_asset::ConcreteFungibleAsset;
-pub use nft::NFT;
+pub use nft::*;
```
shaunxw, line 22: keep it `pub use nft::NFT;` to be consistent with other imports

traits/src/nft.rs
xlc: In that case do we need `balance`? And should just rename this to something like `InspectExtended`

syan095: yeah good idea. I'll look into this

syan095: There's no use case for "balance" right now, but I kept it in case if it is needed in the future.

xlc (approved):

Issue #665, opened 2021-12-08
Would an escrow system fit in this repository?


olanod:
In Virto Network we base our multi currency system on orml traits and tokens pallet, on top of which we built our [`payments-pallet`](https://github.com/virto-network/virto-node/blob/master/pallets/payment), which provides an escrow like system that allows merchants receive payments that won't be fully available until the customer decides some off-chain condition was met and proceeds to release the frozen funds. 

For Virto this is a crucial piece of functionality in our mission to provide easy to use APIs for decentralized commerce applications, specially our initial star use case of decentralized on&off ramps for the ecosystem. That being said the payments pallet remains a very generic and useful primitive that I'm sure others could benefit from and we would be glad to contribute back to the ecosystem with the added benefit that more eyes can translate in higher quality code.

xlc:
We are always welcome to any generalized reusable high quality pallets.

You can create some basic design and share it to the parachain teams to see if there is indeed an interest of such pallet.

PR #691, opened 2022-02-08
Payments pallet

olanod:
[virto]: https://virto.network
[grant]: https://github.com/w3f/Grants-Program/blob/master/applications/lip_payments.md
[pallet]: https://github.com/virto-network/virto-node/tree/master/pallets/payment

Continuing the discussion in https://github.com/open-web3-stack/open-runtime-module-library/issues/665. We, the [Virto Network][virto] team recently finished our first [Web3 grant][grant] milestone for the development of the [`payments_pallet`][pallet], a flexible and economically secure escrow-like system that allows people pay for off-chain products or services with any given `MultiReservableCurrency`, we believe our pallet can have numerous use cases outside our parachain and being part of the ORML code base makes it more reachable to the community.

### Main Features
- [x] **Basic escrow-like functionality:** when users make a payment funds are transferred to destination account but remain on hold until the user triggers a `release`(usually when an off-chain event like a bank transfer has taken place)
- [x] **Dispute resolution:** Each payment is assigned a configurable _judge_ account that can resolve disputes when the parties disagree.
- [x] **Fee handler:** Users of the pallet can implement their own logic to charge extra fees per payment.
- [x] **Payment remarks:** Payments can optionally include a "remark" that can give external components some context about what's being payed(e.g. the fee handler can use that info to charge fees based on the remark data)
- [x] **Payment request:** Recipients can initiate payments that just need to be accepted by the payer. _*Not part of Web3 grant_
~**Recurrent payments:** Ability to create payments that repeat periodically~ _TBD. *Not part of Web3 grant_


payments/src/lib.rs
```diff
@@ -0,0 +1,578 @@
...
+				T::IncentivePercentage::get(),
+				Some(remark),
+			)?;
+			// reserve funds for payment
+			<Self as PaymentHandler<T>>::reserve_payment_amount(&who, &recipient, payment_detail)?;
+			// emit paymentcreated event
+			Self::deposit_event(Event::PaymentCreated {
+				from: who,
+				asset,
+				amount,
```
xlc, line 196: The remark should be in the event and I don't see too much reason to have two pay. We can always use the with remark version and pass empty remark if it is not needed.

payments/src/lib.rs
```diff
@@ -0,0 +1,578 @@
...
+		}
+
+		/// Cancel a payment in created state, this will release the reserved
+		/// back to creator of the payment. This extrinsic can only be called by
+		/// the recipient of the payment
+		#[transactional]
+		#[pallet::weight(T::WeightInfo::cancel())]
+		pub fn cancel(origin: OriginFor<T>, creator: T::AccountId) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+			if let Some(payment) = Payment::<T>::get(creator.clone(), who.clone()) {
```
xlc, line 227:
```suggestion
			if let Some(payment) = Payment::<T>::get(&creator, &who) {
```

payments/src/lib.rs
```diff
@@ -0,0 +1,578 @@
...
+							Percent::from_percent(0),
+						)?;
+						Self::deposit_event(Event::PaymentCancelled { from: creator, to: who });
+					}
+					// if the payment is in state PaymentRequested, remove from storage
+					PaymentState::PaymentRequested => Payment::<T>::remove(creator.clone(), who.clone()),
+					_ => fail!(Error::<T>::InvalidAction),
+				}
+			} else {
+				fail!(Error::<T>::InvalidPayment);
```
xlc, line 243: why not return it

payments/src/lib.rs
```diff
@@ -0,0 +1,578 @@
...
+					PaymentState::Created => {
+						<Self as PaymentHandler<T>>::settle_payment(
+							creator.clone(),
+							who.clone(),
+							Percent::from_percent(0),
+						)?;
+						Self::deposit_event(Event::PaymentCancelled { from: creator, to: who });
+					}
+					// if the payment is in state PaymentRequested, remove from storage
+					PaymentState::PaymentRequested => Payment::<T>::remove(creator.clone(), who.clone()),
```
xlc, line 239:
```suggestion
					PaymentState::PaymentRequested => Payment::<T>::remove(&creator, &who),
```

xlc (commented):
There should be some minimal payment amount to prevent spam

olanod:
> There should be some minimal payment amount to prevent spam

@xlc wouldn't the usual system fee calculated from the weight be enough to protect from spam? It's hard to determine what a reasonable minimum payment amount should be since different tokens can differ greatly in value, but I assume we want to at least ensure the receiving account gets the minimum existential deposit to not reaped? Should [`MultiCurrency::minimum_balance`](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/traits/src/currency.rs#L42) do the trick? Is the current tokens implementation using that minimum balance to keep the account alive?

xlc:
The fee is not enough. It only covers the computation, not storage.

The principle is that every storage write must come with some form of spam protection. It can be a fixed amount deposit (like pallet proxy does), or some valuable assets is locked (ED). 

ED is low and can only cover so much storage usage so the protection will be weak if you expand ED to cover too much storages and this is just an additional expansion.

Ultimately it should be the runtime to decide the ED value, how much storages it can cover, and if this particular feature is included or now. You shouldn't avoid making decision for the runtime within pallet, which will make it less reusable.

The the simple solution is to expose a hook to check if payment amount so you can defer the decision to the runtime. Then you can simply check against with `minimum_balance` if you want to. Others could figure out the right number for them.

olanod:
Understood, reserving an extra deposit for the storage cost of the remark which is our main variable is in our todo list, I was thinking of making it part of the "incentive amount" to not add yet another fee and configuration parameter that hurts usability, incentive amount is an extra percentage reserved from the payment returned after release(when payment detail is removed from storage), it can remain being a percentage but we can require it to not fall bellow a minimum that depends on the size of the remark.

xlc:
Why do you even need to store the remark onchain? It can be queried easily with an indexer.

Also currently this pallet won't support multiple payments between a sender and receiver, which is maybe fine, but less than ideal.

olanod:
Fair point about the remark we could change that, I wasn't seeing it like the system remark but a more important part of the payment detail that other pallets could access for whatever reason.  In Virto we plan to use it within the `FeeHandler`, we have the concept of local communities that will collect a tax and marketplaces that are a kind of token curated registry that also get a percentage of the payment, p2p payments outside a marketplace will pay a higher fee that goes to the local community and payments via a marketplace pay a smaller fee(a tiny part still going to the local community) where we use the remark to know the marketplace and potentially the kind of article in case some items in the future will be taxed differently. Since the remark is still in scope while doing the fees calculation(apply_fees will return a list of account->fee) I guess we don't actually need to store it.

The single payment per sender-receiver is actually on purpose, it makes the pallet simpler and it proved better for users actually since its easier to know that you only have one transaction at the time with a given client, at least for p2p interactions that we handle in a chat is simpler and more than acceptable, imagine the treasury paying a team for some development, you usually want them to finish one job before staring another. In the future it might become a bigger limitation but I guess we'll let users be the judges of that :)

stanly-johnson:
Not sure why the CI fails, `make dev-format-check` passes without issue on local machine

xlc:
> Not sure why the CI fails, `make dev-format-check` passes without issue on local machine

That will be due to rust version. Please make sure you are using a later version https://github.com/open-web3-stack/open-runtime-module-library/blob/9e041dc9d213f843b18b3008f32f3acabb287dcb/.github/workflows/test.yml#L36

olanod:
Hi @xlc sorry to bother but it seems like you are the only one who can currently help us with this review.  
In a related note, a topic for a [Substrate seminar](https://github.com/substrate-developer-hub/substrate-seminar/issues/8) popped up about diving into ORML and how to contribute to it, it would be interesting to use this chance to think of ways ORML could be made more community driven and have better contribution guidelines in place that can help bring good quality content to the repository. Would love to hear some thoughts on the subject.

payments/src/lib.rs
```diff
@@ -0,0 +1,601 @@
+#![allow(clippy::unused_unit, unused_qualifications, missing_debug_implementations)]
```
xlc, line 1: can you add crate docs, just copy the readme will be good

xlc: are those allow absolutely necessary? some comments to explain them will be helpful

payments/src/lib.rs
```diff
@@ -0,0 +1,601 @@
...
+			// reduce the weight used to read the task list
+			remaining_weight = remaining_weight.saturating_sub(T::WeightInfo::remove_task());
+
+			ScheduledTasks::<T>::mutate(|tasks| {
+				let mut task_list: Vec<_> = tasks
+					.clone()
+					.into_iter()
+					// leave out tasks in the future
+					.filter(|(_, ScheduledTask { when, task })| when <= &now && matches!(task, Task::Cancel))
+					.collect();
```
xlc, line 173: should limit number of tasks to take

payments/src/lib.rs
```diff
@@ -0,0 +1,601 @@
...
+
+				// order by oldest task to process
+				task_list.sort_by(|(_, t), (_, x)| x.when.partial_cmp(&t.when).unwrap());
```
xlc, line 176: why use partial_cmp & unwrap instead of cmp?
also depends on number of expected tasks, we may want to change the data structure to have the tasks presorted

payments/src/lib.rs
```diff
@@ -0,0 +1,601 @@
...
+
+		/// Release any created payment, this will transfer the reserved amount
+		/// from the creator of the payment to the assigned recipient
+		#[transactional]
+		#[pallet::weight(T::WeightInfo::release())]
+		pub fn release(origin: OriginFor<T>, to: T::AccountId) -> DispatchResultWithPostInfo {
+			let from = ensure_signed(origin)?;
+
+			// ensure the payment is in Created state
+			if let Some(payment) = Payment::<T>::get(&from, &to) {
```
xlc, line 262:
```suggestion
			let payment = Payment::<T>::get(&from, &to).ok_or(Error::<T>::InvalidPayment)?;
```

payments/src/types.rs
```diff
@@ -0,0 +1,122 @@
...
+	Cancel,
+}
+
+/// The details of a scheduled task
+#[derive(PartialEq, Eq, Clone, Encode, Decode, Debug, TypeInfo, MaxEncodedLen)]
+pub struct ScheduledTask<Time: HasCompact> {
+	/// the type of scheduled task
+	pub task: Task,
+	/// the 'time' at which the task should be executed
+	#[codec(compact)]
```
xlc, line 120: no need to make this compact

payments/src/lib.rs
stanly-johnson: The DS to use here is tricky, if we use a sorted BoundedVec<>, the `dispute_refund` extrinsic will get expensive, hence the use of map here.

stanly-johnson: The DS to use here is tricky, if we use a sorted BoundedVec<>, the `dispute_refund` extrinsic will get expensive, hence the use of map here.

payments/src/types.rs
stanly-johnson: https://github.com/open-web3-stack/open-runtime-module-library/pull/691/commits/0377671b869fca260c25e243b0a32c9b44cde5f1

xlc:
Can you upgrade to polkadot-v0.9.19 and address my comments and then it should be good to merge

payments/src/lib.rs
stanly-johnson: https://github.com/open-web3-stack/open-runtime-module-library/pull/691/commits/16be33026fddbef327e030a93f8cd190f6491366

xlc: this commit is lost? also better to use `.take(n)` than truncate

xlc (approved):

PR #699, opened 2022-02-20
add parachain c

zqhxuyuan:
some parachain may not use `GeneralKey` as its token, but use just its parachain, i.e. PHA:

https://github.com/AcalaNetwork/Acala/blob/d92b2db7e384ebf24dafc89d7cc0e04ee4724435/runtime/karura/src/lib.rs#L1811-L1812

```rust
// Phala Native token
Token(PHA) => Some(MultiLocation::new(1, X1(Parachain(parachains::phala::ID)))),
```

this PR add parachain c, and use parachain(3) to represent its token.

### DIFF
--- a/xtokens/src/mock/mod.rs
+++ b/xtokens/src/mock/mod.rs
@@ -30,7 +30,7 @@ pub enum CurrencyId {
 	/// Parachain B B1 token
 	B1,
 	/// Parachain C native token
-	C
+	C,
 }
 
 pub struct CurrencyIdConvert;
--- a/xtokens/src/mock/mod.rs
+++ b/xtokens/src/mock/mod.rs
@@ -29,6 +29,8 @@ pub enum CurrencyId {
 	B,
 	/// Parachain B B1 token
 	B1,
+	/// Parachain C native token
+	C
 }
 
 pub struct CurrencyIdConvert;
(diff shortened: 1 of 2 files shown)


xtokens/src/mock/mod.rs
```diff
@@ -58,13 +61,15 @@ impl Convert<MultiLocation, Option<CurrencyId>> for CurrencyIdConvert {
...
 				X2(Parachain(2), GeneralKey(k)) if k == b1 => Some(CurrencyId::B1),
+				X1(Parachain(3)) => Some(CurrencyId::C),
 				_ => None,
 			},
 			MultiLocation { parents, interior } if parents == 0 => match interior {
 				X1(GeneralKey(k)) if k == a => Some(CurrencyId::A),
 				X1(GeneralKey(k)) if k == b => Some(CurrencyId::B),
 				X1(GeneralKey(k)) if k == a1 => Some(CurrencyId::A1),
 				X1(GeneralKey(k)) if k == b1 => Some(CurrencyId::B1),
+				Here => Some(CurrencyId::C),
```
shaunxw, line 72: Parachains in mock shares the same `CurrencyIdConvert`, this will make all of them convert `Here` to `CurrencyId::C`, which doesn't make sense.

You'll need to define a convert impl only for Parachain `C`.

Issue #698, opened 2022-02-19
support using KSM as fee(cases: transfer back asset to Statemine)


zqhxuyuan:
current, we have multi assets support, and use `fee_item` asset as reserve

https://github.com/open-web3-stack/open-runtime-module-library/blob/4a66b299037cc3997689538f82847785f9afa65d/xtokens/src/lib.rs#L590

i.e. if `currencies=vec![(CurrencyId::R, 100), (CurrencyId::B, 450)]`, and fee_item=0, reserve=Parent, because fee asset is R, then xcm sent to Relaychain likes:
```
origin: MultiLocation { parents: 0, interior: X1(Parachain(1)) }, 
message: Xcm([
    WithdrawAsset(MultiAssets([
        MultiAsset { id: Concrete(MultiLocation { parents: 0, interior: Here }), fun: Fungible(100) }, 
        MultiAsset { id: Concrete(MultiLocation { parents: 0, interior: X2(Parachain(2), GeneralKey([66])) }), fun: Fungible(450) }])), // 🔥
    ClearOrigin, 
    BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 0, interior: Here }), fun: Fungible(50) }, weight_limit: Limited(40) }, 
    DepositReserveAsset { 
        assets: Wild(All), max_assets: 2, 
        dest: MultiLocation { parents: 0, interior: X1(Parachain(2)) }, 
        xcm: Xcm([
            BuyExecution { fees: MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(50) }, weight_limit: Limited(40) }, 
            DepositAsset { assets: Wild(All), max_assets: 2, beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: Any, id: [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1] }) } }]) }
]), weight_limit: 18446744073709551615, weight_credit: 0
```
this will error when execute second WithdrawAsset because relaychain doesn't have `(0, Parachain(2), GeneralKey(B))` asset.

---

[...]

PR #700, opened 2022-02-22
use relaychain asset as fee

zqhxuyuan:
checkout context infomation on: https://github.com/open-web3-stack/open-runtime-module-library/issues/698

~~for not breaking current semantic([don't allow different reserve](https://github.com/open-web3-stack/open-runtime-module-library/blob/4a66b299037cc3997689538f82847785f9afa65d/xtokens/src/lib.rs#L584-L586)), I use another disptach call instead mixed in `do_transfer_multiassets`. (i.e. may be need another paramether `allow_different_reserve` to implemented.)~~

~~the `transfer_using_relaychain_as_fee` only used for parachain(A) send token to sibling parachain(B), and use relaychain token(R) as fee. thus, the reserve location is decided by sibling asset instead of fee asset.~~

Edit:
for the case of transfer custom asset(RMRK) from Karura to Statemine, we current use two xcm implementation: 
- first one send to relaychain and route to Statemine(Karura-[KSM]-Statemine NonReserve case), transfer user's fee to parachain_2000 sovereign account on Statemine. 
- second one direct send to Statemine(Karura-[KSM,RMRK]-Statemine, ToReserve case)

~~we also has the case of transfer RMRK from Karura to Bifrost. this case also need use KSM as fee too.~~

acala statemine tests: https://github.com/AcalaNetwork/Acala/pull/1712


xtokens/src/lib.rs
```diff
@@ -394,6 +394,53 @@ pub mod module {
...
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			let dest: MultiLocation = (*dest).try_into().map_err(|()| Error::<T>::BadVersion)?;
+
+			let location: MultiLocation = T::CurrencyIdConvert::convert(currency_id.clone())
+				.ok_or(Error::<T>::NotCrossChainTransferableCurrency)?;
+			let asset: MultiAsset = (location, amount.into()).into();
+
+			let fee = MultiAsset {
+				id: AssetId::Concrete(MultiLocation::new(1, Junctions::Here)),
```
shaunxw, line 416: `orml-xtokens` shouldn't assume the location of a particular asset. Make it a type in config.

xtokens/src/lib.rs
```diff
@@ -394,6 +394,53 @@ pub mod module {
...
+				fun: Fungibility::Fungible(dest_weight.into()),
+			};
+
+			let mut assets = MultiAssets::new();
+			assets.push(asset.clone());
+			assets.push(fee.clone());
+
+			let (_, dest, reserve, recipient) = Self::transfer_kind(&asset, &dest)?;
+			ensure!(dest == reserve, "Asset should match to destination!");
```
shaunxw, line 425: We don't use string for dispatchable call errors. Define a variant in `Error`.

xtokens/src/lib.rs
```diff
@@ -394,6 +394,53 @@ pub mod module {
 
 			Self::do_transfer_multiassets(who, assets.clone(), fee.clone(), dest, dest_weight, true)
 		}
+
+		/// Transfer currency and use relay-chain asset as fee
```
shaunxw, line 398: Need more descriptions about which edge case this call handles. Refer to docstrings in other dispatchable calls.

xtokens/src/mock/para.rs
```diff
@@ -266,7 +266,7 @@ impl Convert<AccountId, MultiLocation> for AccountIdToMultiLocation {
 
 parameter_types! {
 	pub SelfLocation: MultiLocation = MultiLocation::new(1, X1(Parachain(ParachainInfo::get().into())));
-	pub const BaseXcmWeight: Weight = 100_000_000;
+	pub const BaseXcmWeight: Weight = 10;
```
shaunxw, line 269: Any reason changing this?

shaunxw (commented):
I would recommend to update `transfer_with_fee` impl instead of adding a new dispatchable call.

The new `transfer_using_relaychain_as_fee` is basically the same with `do_transfer_multiassets`, but it doesn't require fee and other assets share the same reserve, plus it use non-fee assets as reserve.

Updating `do_transfer_multiassets` and make it
1. Only ensure non-fee assets share the same reserve location.
2. Refer to non-fee assets for reserve location.

should work. As it only extends use cases of it with less limitations, it's not a breaking change for `transfer_with_fee`.

zqhxuyuan:
> Only ensure non-fee assets share the same reserve location.
Refer to non-fee assets for reserve location.

That make sense, but I think change `transfer_with_fee` will break existing case,  i.e.
https://github.com/open-web3-stack/open-runtime-module-library/blob/f8f6f82fc08c47ead0346a215a941be75fdc8fd3/xtokens/src/tests.rs#L310-L315

`transfer_with_fee` only support one asset and use that asset as fee. If we want to use relaychain asset as fee and transfer another asset, I think we need to change `transfer_multiasset_with_fee` or `transfer_multicurrencies` or `transfer_multiassets`.

shaunxw:
> That make sense, but I think change transfer_with_fee will break existing case, i.e.

My bad, I mean the new added calls which can specify fee asset type. Updating `do_transfer_multiassets` impl should work, only removing the fee and other assets limitations mentioned above shouldn't break any existing call.

xtokens/src/mock/para.rs
zqhxuyuan: reverted

xtokens/src/lib.rs
zqhxuyuan: changes in `do_transfer_multiassets `

xtokens/src/lib.rs
zqhxuyuan: changes in `do_transfer_multiassets `

xtokens/src/lib.rs
zqhxuyuan: changes in `do_transfer_multiassets `

shaunxw (commented):
`DistinctReserveForAssetAndFee` error variant should be removed, as there is not such limitation any more.

zqhxuyuan:
I think `DistinctReserveForAssetAndFee` still needed, because there maybe case like `[R,A,A1]` which is valid, but `[R,A,B]` is invalid.

xtokens/src/lib.rs
```diff
@@ -571,34 +570,121 @@ pub mod module {
 				assets.len() <= T::MaxAssetsForTransfer::get(),
 				Error::<T>::TooManyAssetsBeingSent
 			);
+			let origin_location = T::AccountIdToMultiLocation::convert(who.clone());
+
+			let fee_to_dest = reset_fee(&fee, dest_weight as u128);
```
xlc, line 575: `dest_weight` is weight not the amount?

xtokens/src/lib.rs
```diff
@@ -571,34 +570,121 @@ pub mod module {
...
+			let mut non_fee_reserve: Option<MultiLocation> = None;
+			let asset_len = assets.len();
+			let mut assets_to_dest = MultiAssets::new();
+			for i in 0..asset_len {
 				let asset = assets.get(i).ok_or(Error::<T>::AssetIndexNonExistent)?;
 				if !asset.is_fungible(None) {
 					return Err(Error::<T>::NotFungible.into());
 				}
 				if fungible_amount(asset).is_zero() {
 					return Ok(());
```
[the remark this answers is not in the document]
zqhxuyuan: I think previous we don't allow asset that is zero amount. it's reasonable that user should input non zero amount.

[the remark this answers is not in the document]
zqhxuyuan: I checked the `fungible_amount`, only support `Fungible` for now. asset that is not `Fungible` is zero.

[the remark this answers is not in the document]
xlc: Yes but the issue is that we can have multiple fungible assets. If one of it is zero and one is not, this currently will return Ok and not doing any transfer, which is a bug.
It should skip the zero one, not return the method.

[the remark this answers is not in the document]
xlc: That's a bug outside of this PR. I will make another PR to fix it.

xtokens/src/lib.rs
zqhxuyuan: the `dest_weight` should be convert to fee amount, if using `WeightToFee` trait, it only works on one parachain. so here add another config to xTokens module: `MinXcmFee` which can config many parachain(kind like of `ExistentialDeposits`). parachain don't want to this feature can simply config as `()`.

xtokens/src/lib.rs
```diff
@@ -83,6 +83,9 @@ pub mod module {
 		#[pallet::constant]
 		type SelfLocation: Get<MultiLocation>;
 
+		/// Minimum xcm execution fee paid on destination chain.
+		type MinXcmFee: GetOptionValueByKey<MultiLocation, u128>;
```
xlc, line 87:
```suggestion
		type MinXcmFee: GetValueByKey<MultiLocation, Option<u128>>;
```

I don't think we need a version for option

xtokens/src/lib.rs
```diff
@@ -467,43 +470,125 @@ pub mod module {
...
-					Self::transfer_to_reserve(assets.clone(), fee.clone(), dest.clone(), recipient, dest_weight)?
+			// In case that fee reserve != asset reserve, there're two xcm sent from sender.
+			// first xcm send to fee reserve which also route to dest. second xcm directly
+			// send to dest. the fee amount in fee asset is split into two parts.
+			// 1. assets send to fee reserve = fee_amount - dest_weight
+			// 2. assets send to dest reserve = dest_weight
+			// the first part + second part = fee amount in fee asset
+			let fee_reserve = fee.reserve();
+			if fee_reserve != non_fee_reserve {
+				ensure!(non_fee_reserve.is_some(), Error::<T>::InvalidDest);
```
xlc, line 504: this check is redundant, covered by the next check

xtokens/src/lib.rs
```diff
@@ -467,43 +469,140 @@ pub mod module {
...
+			// sender. first xcm sent to fee reserve chain which route to dest chain. second
+			// xcm directly sent to dest chain. the fee amount in fee asset is split into
+			// two parts. 1. fee asset sent to fee reserve chain = fee_amount - min_xcm_fee
+			// 2. fee asset sent to dest reserve chain = min_xcm_fee
+			// the first part + second part = fee amount in fee asset
+			let fee_reserve = fee.reserve();
+			if fee_reserve != non_fee_reserve {
+				// Current only support `ToReserve` with relay-chain asset as fee. other case
+				// like `NonReserve` or `SelfReserve` with relay-chain fee is not support.
+				ensure!(non_fee_reserve == dest.chain_part(), Error::<T>::InvalidAsset);
```
zqhxuyuan, line 505: should support NonReserve, i.e. Karura-[RMRK, KSM(fee)]->Bifrost

xtokens/src/lib.rs
```diff
@@ -467,43 +469,140 @@ pub mod module {
...
+				let asset_to_fee_reserve = subtract_fee(&fee, min_xcm_fee);
+				assets_to_fee_reserve.push(asset_to_fee_reserve.clone());
+
+				// First xcm sent to fee reserve chain and routed to dest chain.
+				// The `SelfLocation` current is (1, Parachain(id)) refer to sender parachain.
+				// we use `SelfLocation` to fund fee to sender's parachain sovereign account on
+				// destination chain, which asset is originated from sender account on sender
+				// chain. This means if user setup too much fee, the fee is not returned to
+				// user, instead deposit to sibling parachain sovereign account on dest chain.
+				// Notice: if parachain set `SelfLocation` to (0, Here), it'll be error!
```
shaunxw, line 534: This should be documented in `SelfLocation`.

xtokens/src/lib.rs
```diff
@@ -467,43 +469,140 @@ pub mod module {
...
+				)?;
+
+				// Second xcm send to dest chain.
+				// Current not ensure xcm order delivery. if second xcm is executed before first
+				// xcm, then second xcm may failed because of sibling parachain account don't
+				// have enough fee to withdraw. we can pre-fund some amount to sibling parachain
+				// sovereign account to fix this issue. when first xcm executed later on, the
+				// sibling sovereign parachain account is deposit. and next transaction will
+				// succeed even though second xcm is executed before first xcm if user fee is
+				// less than parachain sovereign account balance.
```
shaunxw, line 552: Dispatchable calls docstring or pallet readme should include these usage notes.

xtokens/src/lib.rs
```diff
@@ -813,3 +920,18 @@ fn half(asset: &MultiAsset) -> MultiAsset {
...
 	}
 }
+
+fn subtract_fee(asset: &MultiAsset, amount: u128) -> MultiAsset {
+	let final_amount = fungible_amount(asset).checked_sub(amount).expect("fee too low; qed");
+	MultiAsset {
+		fun: Fungible(final_amount),
+		id: asset.id.clone(),
+	}
+}
```
shaunxw, line 930: `(final_amount, asset.id.clone()).into()` should work

xtokens/src/lib.rs
```diff
@@ -467,43 +469,140 @@ pub mod module {
...
+		}
+
+		/// Send xcm with given assets and fee to dest or reserve chain.
+		fn send_xcm(
+			origin_location: MultiLocation,
+			assets: MultiAssets,
+			fee: MultiAsset,
+			reserve: Option<MultiLocation>,
+			dest: &MultiLocation,
+			manual_recipient: Option<MultiLocation>,
```
shaunxw, line 591:
```suggestion
			maybe_recipient_override: Option<MultiLocation>,
```

xtokens/src/lib.rs
```diff
@@ -467,43 +469,140 @@ pub mod module {
...
+				assets,
+				fee,
+				dest,
+			});
+
+			Ok(())
+		}
+
+		/// Send xcm with given assets and fee to dest or reserve chain.
+		fn send_xcm(
```
shaunxw, line 585: need a better naming

xtokens/src/lib.rs
```diff
@@ -813,3 +920,18 @@ fn half(asset: &MultiAsset) -> MultiAsset {
...
+		id: asset.id.clone(),
+	}
+}
+
+fn reset_fee(asset: &MultiAsset, amount: u128) -> MultiAsset {
+	MultiAsset {
+		fun: Fungible(amount),
+		id: asset.id.clone(),
+	}
+}
```
shaunxw, line 937: Overkill to add a function to build a `MultiAsset`.

xtokens/src/lib.rs
```diff
@@ -813,3 +920,18 @@ fn half(asset: &MultiAsset) -> MultiAsset {
 		id: asset.id.clone(),
 	}
 }
+
+fn subtract_fee(asset: &MultiAsset, amount: u128) -> MultiAsset {
+	let final_amount = fungible_amount(asset).checked_sub(amount).expect("fee too low; qed");
```
shaunxw, line 925: There is no `fee_amount >= min_xcm_fee` check. This could panic no?

xtokens/src/lib.rs
zqhxuyuan: No support `NonReserve` for now, user can first transfer RMRK from Karura to Statemine, then transfer RMRK from Statemine to Bifrost.

xtokens/src/lib.rs
zqhxuyuan: rename to `execute_and_send_reserve_kind_xcm`

xtokens/src/lib.rs
zqhxuyuan: we checked before subtract, so it's ok here I think
```
// min xcm fee should less than user fee
let fee_to_dest: MultiAsset = (fee.id.clone(), min_xcm_fee).into();
ensure!(fee_to_dest < fee, Error::<T>::InvalidAsset);
```

zqhxuyuan: maybe we need add another Error type such as `FeeNotEnough`

xtokens/src/lib.rs
zqhxuyuan: move to README

xtokens/src/lib.rs
zqhxuyuan: move to README

shaunxw (approved):

PR #702, opened 2022-02-22
Add a test for cliff vesting

cowboy-bebug:
Hi team,

We're using the `orml-vesting` pallet for both linear and cliff vesting and thought it would be nice to add a test for it (so that other developers can make an informative decision on this pallet compared to `pallet-vesting` which lacks the support for cliff).

Thanks! ❤️

### DIFF
--- a/vesting/src/tests.rs
+++ b/vesting/src/tests.rs
@@ -413,3 +413,45 @@ fn exceeding_maximum_schedules_should_fail() {
 		);
 	});
 }
+
+#[test]
+fn cliff_vesting_works() {
+	const VESTING_AMOUNT: u64 = 12;
+	const VESTING_PERIOD: u64 = 20;
+
+	ExtBuilder::build().execute_with(|| {
+		let cliff_schedule = VestingSchedule {
+			start: VESTING_PERIOD - 1,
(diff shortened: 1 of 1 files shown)


vesting/src/tests.rs
```diff
@@ -413,3 +413,45 @@ fn exceeding_maximum_schedules_should_fail() {
...
+			amount: VESTING_AMOUNT,
+			reasons: Reasons::All,
+		};
+
+		assert_eq!(PalletBalances::free_balance(BOB), 0);
+		assert_ok!(Vesting::vested_transfer(Origin::signed(ALICE), BOB, cliff_schedule));
+		assert_eq!(PalletBalances::free_balance(BOB), VESTING_AMOUNT);
+		assert_eq!(PalletBalances::locks(&BOB), vec![balance_lock.clone()]);
+
+		(1..VESTING_PERIOD).into_iter().for_each(|i| {
```
xlc, line 441: why not just a for loop?

cowboy-bebug: It was just a preference. I'll refactor to more readable for loop

xlc (approved):

PR #706, opened 2022-02-27
XTokens refactor

xlc:
Unified events so there is only one to deal with
Not allow zero amount because apparently it is not allowed https://github.com/paritytech/polkadot/issues/4996

### DIFF
--- a/xtokens/src/tests.rs
+++ b/xtokens/src/tests.rs
@@ -847,7 +847,7 @@ fn send_with_zero_fee_should_yield_an_error() {
 				),
 				40,
 			),
-			Error::<para::Runtime>::FeeCannotBeZero
+			Error::<para::Runtime>::ZeroFee
 		);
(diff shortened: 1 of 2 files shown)


xtokens/src/tests.rs
```diff
@@ -1048,3 +1048,54 @@ fn specifying_a_non_existent_asset_index_should_fail() {
...
+					)
+						.into()
+				),
+				40,
+			),
+			Error::<para::Runtime>::ZeroAmount
+		);
+	});
+
+	// TODO: should have more tests after https://github.com/paritytech/polkadot/issues/4996
```
xlc, line 1100: should have more test to ensure MultiAsset is valid (fungible & non zero amount) after https://github.com/paritytech/polkadot/issues/4996

shaunxw (approved):
LGTM

PR #708, opened 2022-03-01
Girazoki configurable reserve provider

girazoki:
This is a proposal to configure the way the reserve part is calculated for a MultiAsset in xtokens. It includes adding an associated type called `ReserveProvider`, that chains can configure the way they prefer. For instance, while Acala might be willing to keep the absolute view for Self Tokens, while other chains can implement this the way they want, e.g., changing it to a relative view.

This is just a proposal that can be dropped in case the suggestion is not well received. I understand that this implies some changes in the runtimes importing xtokens, but minimal changes to the xtokens code/tests itself

### DIFF
--- a/traits/src/location.rs
+++ b/traits/src/location.rs
@@ -65,16 +65,10 @@ pub struct RelativeReserveProvider;
 impl Reserve for RelativeReserveProvider {
 	fn reserve(asset: &MultiAsset) -> Option<MultiLocation> {
 		if let Concrete(location) = &asset.id {
-			match (location.parents, location.first_interior()) {
-				// sibling parachain
-				(1, Some(Parachain(id))) => Some(MultiLocation::new(1, X1(Parachain(*id)))),
-				// parent
-				(1, _) => Some(MultiLocation::parent()),
-				// children parachain
-				(0, Some(Parachain(id))) => Some(MultiLocation::new(0, X1(Parachain(*id)))),
-				// All asset that is not a children parachain is a self reserved asset
-				(0, _) => Some(MultiLocation::here()),
-				_ => None,
+			if location.parents == 0 && !is_chain_junction(location.first_interior()) {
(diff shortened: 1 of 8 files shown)


shaunxw (commented):
I'm not against the direction to make absolute/relative location flexible, though a bit concerned that this might not be able to bring actual benefits for cross-chain transfer abilities, but just to meet different parachains developers' preferences. From code maintenance point of view, I would prefer to keep it simple.

The reserve provider needs more tests coverages. Mocks in xtokens should be updated to include both absolute and relative provider in different parachains.

traits/src/location.rs
```diff
@@ -42,19 +42,46 @@ impl Parse for MultiLocation {
...
+			match (location.parents, location.first_interior()) {
+				// sibling parachain
+				(1, Some(Parachain(id))) => Some(MultiLocation::new(1, X1(Parachain(*id)))),
+				// parent
+				(1, _) => Some(MultiLocation::parent()),
+				// children parachain
+				(0, Some(Parachain(id))) => Some(MultiLocation::new(0, X1(Parachain(*id)))),
+				// All asset that is not a children parachain is a self reserved asset
+				(0, _) => Some(MultiLocation::here()),
+				_ => None,
```
zqhxuyuan, line 77: this snippet can be optimized to:

```
			if location.parents == 0 && !is_chain_junction(location.first_interior()) {
				Some(MultiLocation::here())
			} else {
				location.chain_part()
			}
```

shaunxw (commented):
LGTM

Could you please also add unit tests to cover the following scenarios?
- Transfer currency `D` from `ParaA` to `ParaB`
- Transfer currency `R` from `ParaA` to `ParaD`

shaunxw (approved):

xlc:
Needs a resolve conflicts and good to merge @girazoki

PR #712, opened 2022-03-11
Return earlier if the too many currencies

Dengjianping:
This extrinsic should return more early if too many currencies are sent.
It's better to check it before execute the `for` loop: 
https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/src/lib.rs#L445

### DIFF
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -472,6 +472,10 @@ pub mod module {
 			dest: MultiLocation,
 			dest_weight: Weight,
 		) -> DispatchResult {
+			ensure!(
+				assets.len() <= T::MaxAssetsForTransfer::get(),
+				Error::<T>::TooManyAssetsBeingSent
+			);
 			let origin_location = T::AccountIdToMultiLocation::convert(who.clone());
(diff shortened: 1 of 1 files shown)


xtokens/src/lib.rs
```diff
@@ -467,10 +472,6 @@ pub mod module {
 			dest: MultiLocation,
 			dest_weight: Weight,
 		) -> DispatchResult {
-			ensure!(
```
xlc, line 470: we still need this check

Dengjianping: I got your point, because `do_transfer_multiassets` is called by many places, so we cannot simply skip this check, right?.
If you think it's necessary, I will close the pr.

xlc: yes. but I think we are fine to check it twice for the currencies?

Dengjianping: I rollbacked it. Check currencies twice.

zqhxuyuan:
LGTM

xlc (approved):

Issue #721, opened 2022-03-29
Open source asset registry pallet


GopherJ:
what I saw from many parachain teams is that every parachain team build/is building an asset registry pallet for foreign assets, this is not good for polkadot because we are re-inventing the wheel multiple times, altough there maybe gaps to build an asset registry for all but still want to throw my thoughts here

xlc:
The one used by Acala includes a lot of Acala specific features so not suitable as a generic reusable pallet. However it should be possible to extract some of the features into a new pallet to put here.

However I would like to gather more inputs from various teams first.

samelamin:
@GopherJ thank you so much for raising this. Coincidently I asked this very same question on the builder's call yesterday

Totally agree! 

More and more chains will want to implement each other's assets and we need a best practice defined to avoid rework

I suggest we define a clear goal for a pallet that can be imported to your runtime and give you an easy way to import assets from other chains

In addition, we need a guide to ensure future teams implement it in the same way

Essentially something like https://github.com/substrate-developer-hub/substrate-node-template/  but for importing assets. I see a guide has already started which includes a pallet example of an exchange

That helps but as a new chain, my first question would be "how do I allow users on my chain to use $aUSD/$IMBU/$BNC"

Ideally, I would want a clear guide on how to do that 

The teams using ORML can each contribute to it meaning anyone implementing it can easily import all the assets of chains using this pallet 

In an ideal world you can just import the pallet, add/comment out a few lines of code and you now have that asset usable on your chain


I am more than happy contributing to this guide too

GopherJ:
Refs:
- https://github.com/PureStake/moonbeam/tree/master/pallets/asset-manager
- https://github.com/AcalaNetwork/Acala/tree/master/modules/auction-manager
- https://github.com/parallel-finance/parallel/tree/master/pallets/asset-manager

enthusiastmartin:
Hi, 

We in HydraDX / Basilisk also use our own asset-registry pallet. 

https://github.com/galacticcouncil/warehouse/tree/main/asset-registry

Our Asset registry pallet requirements would be as follows. This is basically what our asset-registy does.

- adding new assets dynamically
- auto assign ids to new newly added asset - we use sequential ids ( this might be challenging to do as general approach as other projects might need to explicitly specify asset id )
- different type of assets 
- existential deposit / min balance.
- location of an asset - used for foreign assets where it is required to specify MultiLocation and it is used to resolve location to asset in xcm.
[...]

PR #728, opened 2022-04-21
feat: asset registry

sander2:
This is meant to address https://github.com/open-web3-stack/open-runtime-module-library/issues/721.

This PR will need cleanup, tests, etc, but I'm opening this as a draft now to get some feedback regarding the direction. Some things to note about the implementation:

- In the issue I raised two possible implementations: either we can provide some of the metadata fields by default, or we leave it fully customizable. I went with the latter approach in this PR.
- The only dispatchable function is one to register (or update) an asset. 
- The metadata is fully customizable - the only requirement being that there is a way to extract a MultiAsset from it. This is needed to be able to map xcm assets to assets in this pallet
- The assigned asset id is customizable. The last assigned asset id is stored, so you can either derive a sequential id or base it purely on the metadata
- Some traits and trait implementations are provided to allow for easy WeightTrader implementations. Unfortunately because `WeightTrader::refund_weight` allows only a single asset to be refunded, the default implementation does not support paying with different assets. That is, it can pay with any asset in the registry, but only one asset per xcm message. I opened https://github.com/paritytech/polkadot/issues/5364 to see if `refund_weight` can refund multiple assets in the future. 

An example of usage:
``` rust
// the metadata definition...
#[derive(scale_info::TypeInfo, Encode, Decode, Clone, Eq, PartialEq, Debug)]
pub struct AssetMetadata {
    pub location: MultiLocation,
    pub decimals: u32,
    pub fee_per_second: u128,
    pub name: Vec<u8>,
    pub symbol: Vec<u8>,
}
impl Into<MultiLocation> for AssetMetadata {
    fn into(self) -> MultiLocation {
        self.location
    }
}

// the config...
impl orml_asset_registry::Config for Runtime {
    type Event = Event;
    type AssetMetadata = AssetMetadata;
    type AssetId = u32;
    type AuthorityOrigin = EnsureRoot<AccountId>;
    type AssignAsset = SequentialId<u32>;
}

// for paying xcm fees..
pub struct MyFixedConversionRateProvider;
impl FixedConversionRateProvider for MyFixedConversionRateProvider {
    fn get_fee_per_second(location: &MultiLocation) -> Option<u128> {
        let metadata = AssetRegistry::fetch_metadata_by_location(location)?;
        Some(metadata.fee_per_second)
    }
}
pub type Trader = (
    (...)
    AssetRegistryTrader<FixedRateAssetRegistryTrader<MyFixedConversionRateProvider>, ToTreasury>,
);
```

I would love to get some feedback on this. I'm out of office until tuesday though so might not be able to address any comments until then.

### DIFF
--- a/asset-registry/Cargo.toml
+++ b/asset-registry/Cargo.toml
@@ -1,0 +1,68 @@
+[package]
+name = "orml-asset-registry"
+description = "Registry for (foreign) assets"
+repository = "https://github.com/open-web3-stack/open-runtime-module-library/tree/master/asset-registry"
+license = "Apache-2.0"
+version = "0.4.1-dev"
+authors = ["Interlay Ltd, etc"]
+edition = "2021"
+
+[dependencies]
+scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
+serde = { version = "1.0.136", optional = true }
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["max-encoded-len"] }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20", default-features = false }
+sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20", default-features = false }
+frame-support = {  git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20", default-features = false }
+frame-system = {  git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20", default-features = false }
+orml-traits = { path = "../traits", version = "0.4.1-dev", default-features = false }
+xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20", default-features = false }
+xcm-builder = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20", default-features = false }
+xcm-executor = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20", default-features = false }
+
+[dev-dependencies]
+sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20", default-features = false }
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20" }
+pallet-balances = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.20" }
+
+# cumulus
+cumulus-primitives-core = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.20" }
+cumulus-pallet-dmp-queue = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.20" }
+cumulus-pallet-xcmp-queue = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.20" }
+cumulus-pallet-xcm = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.20" }
+parachain-info = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.20" }
+
+# polkadot
+polkadot-parachain = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20" }
+xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20" }
+xcm-executor = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20" }
+xcm-builder = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20" }
+pallet-xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20" }
+polkadot-runtime-parachains = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20" }
+xcm-simulator = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.20"}
+
+orml-tokens = { path = "../tokens" }
+orml-xtokens = { path = "../xtokens" }
+orml-xcm = { path = "../xcm" }
+orml-xcm-support = { path = "../xcm-support", default-features = false }
+
+[features]
+default = ["std"]
+std = [
+	"serde",
+	"codec/std",
+	"scale-info/std",
+	"sp-runtime/std",
+	"sp-std/std",
+	"frame-support/std",
+	"frame-system/std",
+	"orml-traits/std",
+	"xcm/std",
+	"xcm-builder/std",
+	"xcm-executor/std",
+]
+runtime-benchmarks = [
+	"frame-support/runtime-benchmarks",
+	"frame-system/runtime-benchmarks",
+]
+try-runtime = ["frame-support/try-runtime"]
--- a/asset-registry/README.md
+++ b/asset-registry/README.md
@@ -1,0 +1,10 @@
+# Asset Registry Module
+
+## Overview
+
+This module provides functionality for storing asset metadata. For each asset, it stores the number of decimals, asset name, asset symbol, existential deposit and (optional) location. Additionally, it stores a value of a generic type that chains can use to store any other metadata that the parachain may need (such as the fee rate, for example). It is designed to be easy to integrate into xcm setups. Various default implementations are provided for this purpose.
+
+The pallet only contains two extrinsics, `register_asset` and `update_asset`:
+
+- `register_asset` creates a new asset
+- `update_asset` modifies some (or all) of the fields of an existing asset
--- a/asset-registry/src/impls.rs
+++ b/asset-registry/src/impls.rs
@@ -1,0 +1,169 @@
+use crate::{module::*, AssetMetadata};
+use frame_support::{log, pallet_prelude::*, weights::constants::WEIGHT_PER_SECOND};
+use orml_traits::{
+	asset_registry::{AssetProcessor, FixedConversionRateProvider, WeightToFeeConverter},
+	GetByKey,
+};
+use sp_runtime::{
+	traits::{AtLeast32BitUnsigned, Bounded, CheckedAdd, One},
+	ArithmeticError,
+};
+use sp_std::prelude::*;
+use xcm::v2::prelude::*;
+use xcm_builder::TakeRevenue;
+use xcm_executor::{traits::WeightTrader, Assets};
+
+/// Alias for AssetMetadata to improve readability (and to placate clippy)
+pub type DefaultAssetMetadata<T> = AssetMetadata<<T as Config>::Balance, <T as Config>::CustomMetadata>;
+
+/// An AssetProcessor that assigns a sequential ID
+pub struct SequentialId<T>(PhantomData<T>);
+
+impl<T> AssetProcessor<T::AssetId, DefaultAssetMetadata<T>> for SequentialId<T>
+where
+	T: Config,
+	T::AssetId: AtLeast32BitUnsigned,
+{
+	fn pre_register(
+		id: Option<T::AssetId>,
+		asset_metadata: DefaultAssetMetadata<T>,
+	) -> Result<(T::AssetId, DefaultAssetMetadata<T>), DispatchError> {
+		let next_id = LastAssetId::<T>::get()
+			.checked_add(&T::AssetId::one())
+			.ok_or(ArithmeticError::Overflow)?;
+
+		match id {
+			Some(explicit_id) if explicit_id != next_id => {
+				// we don't allow non-sequential ids
+				Err(Error::<T>::InvalidAssetId.into())
+			}
+			_ => {
+				LastAssetId::<T>::put(&next_id);
+				Ok((next_id, asset_metadata))
+			}
+		}
+	}
+}
+
+/// A default implementation for WeightToFeeConverter that takes a fixed
+/// conversion rate.
+pub struct FixedRateAssetRegistryTrader<P: FixedConversionRateProvider>(PhantomData<P>);
+impl<P: FixedConversionRateProvider> WeightToFeeConverter for FixedRateAssetRegistryTrader<P> {
+	fn convert_weight_to_fee(location: &MultiLocation, weight: Weight) -> Option<u128> {
+		let fee_per_second = P::get_fee_per_second(location)?;
+		let amount = fee_per_second.saturating_mul(weight as u128) / (WEIGHT_PER_SECOND as u128);
+		Some(amount)
+	}
+}
+
+/// Helper struct for the AssetRegistryTrader that stores the data about
+/// bought weight.
+pub struct BoughtWeight {
+	weight: Weight,
+	asset_location: MultiLocation,
+	amount: u128,
+}
+
+/// A WeightTrader implementation that tries to buy weight using a single
+/// currency. It tries all assets in `payment` and uses the first asset that can
+/// cover the weight. This asset is then "locked in" - later calls to
+/// `buy_weight` in the same xcm message only try the same asset.
+/// This is because only a single asset can be refunded due to the return type
+/// of `refund_weight`. This implementation assumes that `WeightToFeeConverter`
+/// implements a linear function, i.e. fee(x) + fee(y) = fee(x+y).
+pub struct AssetRegistryTrader<W: WeightToFeeConverter, R: TakeRevenue> {
+	bought_weight: Option<BoughtWeight>,
+	_phantom: PhantomData<(W, R)>,
+}
+
+impl<W: WeightToFeeConverter, R: TakeRevenue> WeightTrader for AssetRegistryTrader<W, R> {
+	fn new() -> Self {
+		Self {
+			bought_weight: None,
+			_phantom: Default::default(),
+		}
+	}
+
+	fn buy_weight(&mut self, weight: Weight, payment: Assets) -> Result<Assets, XcmError> {
+		log::trace!(
+			target: "xcm::weight",
+			"AssetRegistryTrader::buy_weight weight: {:?}, payment: {:?}",
+			weight, payment,
+		);
+
+		for (asset, _) in payment.fungible.iter() {
+			if let AssetId::Concrete(ref location) = asset {
+				if matches!(self.bought_weight, Some(ref bought) if &bought.asset_location != location) {
+					// we already bought another asset - don't attempt to buy this one since
+					// we won't be able to refund it
+					continue;
+				}
+
+				if let Some(fee_increase) = W::convert_weight_to_fee(location, weight) {
+					if fee_increase == 0 {
+						// if the fee is set very low it could lead to zero fees, in which case
+						// constructing the fee asset item to subtract from payment would fail.
+						// Therefore, provide early exit
+						return Ok(payment);
+					}
+
+					if let Ok(unused) = payment.clone().checked_sub((asset.clone(), fee_increase).into()) {
+						let (existing_weight, existing_fee) = match self.bought_weight {
+							Some(ref x) => (x.weight, x.amount),
+							None => (0, 0),
+						};
+
+						self.bought_weight = Some(BoughtWeight {
+							amount: existing_fee.checked_add(fee_increase).ok_or(XcmError::Overflow)?,
+							weight: existing_weight.checked_add(weight).ok_or(XcmError::Overflow)?,
+							asset_location: location.clone(),
+						});
+						return Ok(unused);
+					}
+				}
+			}
+		}
+		Err(XcmError::TooExpensive)
+	}
+
+	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
+		log::trace!(target: "xcm::weight", "AssetRegistryTrader::refund_weight weight: {:?}", weight);
+
+		match self.bought_weight {
+			Some(ref mut bought) => {
+				let new_weight = bought.weight.saturating_sub(weight);
+				let new_amount = W::convert_weight_to_fee(&bought.asset_location, new_weight)?;
+				let refunded_amount = bought.amount.saturating_sub(new_amount);
+
+				bought.weight = new_weight;
+				bought.amount = new_amount;
+
+				Some((AssetId::Concrete(bought.asset_location.clone()), refunded_amount).into())
+			}
+			None => None, // nothing to refund
+		}
+	}
+}
+
+impl<W: WeightToFeeConverter, R: TakeRevenue> Drop for AssetRegistryTrader<W, R> {
+	fn drop(&mut self) {
+		if let Some(ref bought) = self.bought_weight {
+			R::take_revenue((AssetId::Concrete(bought.asset_location.clone()), bought.amount).into());
+		}
+	}
+}
+
+pub struct ExistentialDeposits<T: Config>(PhantomData<T>);
+
+// Return Existential deposit of an asset. Implementing this trait allows the
+// pallet to be used in the tokens::ExistentialDeposits config item
+impl<T: Config> GetByKey<T::AssetId, T::Balance> for ExistentialDeposits<T> {
+	fn get(k: &T::AssetId) -> T::Balance {
+		if let Some(metadata) = Pallet::<T>::metadata(k) {
+			metadata.existential_deposit
+		} else {
+			// Asset does not exist - not supported
+			T::Balance::max_value()
+		}
+	}
+}
--- a/asset-registry/src/impls.rs
+++ b/asset-registry/src/impls.rs
@@ -4,9 +4,10 @@ use orml_traits::{
 	asset_registry::{AssetProcessor, FixedConversionRateProvider, WeightToFeeConverter},
 	GetByKey,
 };
+use sp_runtime::FixedPointNumber;
 use sp_runtime::{
 	traits::{AtLeast32BitUnsigned, Bounded, CheckedAdd, One},
-	ArithmeticError,
+	ArithmeticError, FixedU128,
 };
 use sp_std::prelude::*;
 use xcm::v2::prelude::*;
@@ -51,7 +52,8 @@ pub struct FixedRateAssetRegistryTrader<P: FixedConversionRateProvider>(PhantomD
 impl<P: FixedConversionRateProvider> WeightToFeeConverter for FixedRateAssetRegistryTrader<P> {
 	fn convert_weight_to_fee(location: &MultiLocation, weight: Weight) -> Option<u128> {
 		let fee_per_second = P::get_fee_per_second(location)?;
-		let amount = fee_per_second.saturating_mul(weight as u128) / (WEIGHT_PER_SECOND as u128);
+		let weight_ratio = FixedU128::saturating_from_rational(weight as u128, WEIGHT_PER_SECOND as u128);
+		let amount = weight_ratio.saturating_mul_int(fee_per_second);
 		Some(amount)
 	}
 }
--- a/asset-registry/src/lib.rs
+++ b/asset-registry/src/lib.rs
@@ -1,0 +1,312 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+// Older clippy versions give a false positive on the expansion of [pallet::call].
+// This is fixed in https://github.com/rust-lang/rust-clippy/issues/8321
+#![allow(clippy::large_enum_variant)]
+
+use frame_support::{pallet_prelude::*, traits::EnsureOrigin, transactional};
+use frame_system::pallet_prelude::*;
+use orml_traits::asset_registry::AssetProcessor;
+use scale_info::TypeInfo;
+use sp_runtime::{
+	traits::{AtLeast32BitUnsigned, Member},
+	DispatchResult,
+};
+use sp_std::prelude::*;
+use xcm::{v2::prelude::*, VersionedMultiLocation};
+
+pub use impls::*;
+pub use module::*;
+pub use weights::WeightInfo;
+
+mod impls;
+mod mock;
+mod tests;
+mod weights;
+
+/// Data describing the asset properties.
+#[derive(scale_info::TypeInfo, Encode, Decode, Clone, Eq, PartialEq, RuntimeDebug)]
+pub struct AssetMetadata<Balance, CustomMetadata: Parameter + Member + TypeInfo> {
+	pub decimals: u32,
+	pub name: Vec<u8>,
+	pub symbol: Vec<u8>,
+	pub existential_deposit: Balance,
+	pub location: Option<VersionedMultiLocation>,
+	pub additional: CustomMetadata,
+}
+
+#[frame_support::pallet]
+pub mod module {
+	use super::*;
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// Additional non-standard metadata to store for each asset
+		type CustomMetadata: Parameter + Member + TypeInfo;
+
+		/// The type used as a unique asset id,
+		type AssetId: Parameter + Member + Default + TypeInfo;
+
+		/// The origin that is allowed to manipulate metadata.
+		type AuthorityOrigin: EnsureOrigin<<Self as frame_system::Config>::Origin>;
+
+		/// A filter ran upon metadata registration that assigns an is and
+		/// potentially modifies the supplied metadata.
+		type AssetProcessor: AssetProcessor<Self::AssetId, AssetMetadata<Self::Balance, Self::CustomMetadata>>;
+
+		/// The balance type.
+		type Balance: Parameter + Member + AtLeast32BitUnsigned + Default + Copy;
+
+		// /// Weight information for extrinsics in this module.
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Asset was not found.
+		AssetNotFound,
+		/// The version of the `VersionedMultiLocation` value used is not able
+		/// to be interpreted.
+		BadVersion,
+		/// The asset id is invalid.
+		InvalidAssetId,
+		/// Another asset was already register with this location.
+		ConflictingLocation,
+		/// Another asset was already register with this asset id.
+		ConflictingAssetId,
+	}
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		RegisteredAsset {
+			asset_id: T::AssetId,
+			metadata: AssetMetadata<T::Balance, T::CustomMetadata>,
+		},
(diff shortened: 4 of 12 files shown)


asset-registry/src/lib.rs
```diff
@@ -0,0 +1,235 @@
...
+	#[pallet::without_storage_info]
+	pub struct Pallet<T>(_);
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(0)]
+		pub fn register_asset(origin: OriginFor<T>, metadata: T::AssetMetadata) -> DispatchResult {
```
xlc, line 107: I think we can add `asset_id: Option<T::AssetId>` parameter so we can specify the asset id if needed.

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,235 @@
...
+	traits::{AtLeast32BitUnsigned, Member},
+	DispatchResult,
+};
+use sp_std::prelude::*;
+use xcm::latest::prelude::*;
+use xcm_builder::TakeRevenue;
+use xcm_executor::{traits::WeightTrader, Assets};
+
+pub trait AssetIdAssigner<AssetId, AssetMetadata> {
+	fn assign_id(asset_metadata: &AssetMetadata, last_id: Option<AssetId>) -> AssetId;
```
xlc, line 20: make it something like `process_asset(id: Option<AssetId>, metadata: AssetMetadata) -> Result<(AssetId, Metadata), DispatchError>` will be more flexible

not really need to pass last id. the implementation can just read it from this pallet

xlc:
I think we should still require some basic fields like name, symbol, decimals. otherwise the js sdk still need specialized code to handle it for different parachains.

asset-registry/src/lib.rs
zqhxuyuan: Agree, parachains like Acala already implements they own asset-registry, If they migrate to orml's asset-registry, seemless migration is helpfull, they could register asset metadata with already exists asset id. In this way, I think `AssetIdAssigner` also needs more trait method for those scene.

enthusiastmartin:
I agree with xlc - that some basic information can be already part of the pallet ( name,symbol, decimals). 

personally, i would not require `Into<MultiLocation>` for asset metadata, and i would not map the location to asset id when asset is registered. 
I'd add ability to set/update/remove location explicitly after asset is already registered.

sander2:
> I think we should still require some basic fields like name, symbol, decimals. otherwise the js sdk still need specialized code to handle it for different parachains.

Sure, I added name, symbol, decimals and multilocation.

> I agree with xlc - that some basic information can be already part of the pallet ( name,symbol, decimals).
> 
> personally, i would not require `Into<MultiLocation>` for asset metadata, and i would not map the location to asset id when asset is registered. I'd add ability to set/update/remove location explicitly after asset is already registered.

When I added default fields, I also added the (multi)location. Are you opposed to this? Do you think it should be an optional item? Or should it be removed as a default field altogether?

asset-registry/src/lib.rs
sander2: Done. However, without the last id, we can't provide a default `SequentialId` implementation - each chain will need to write an explicit implementation now. Because of that, I would actually be in favor of adding it back in as an arg. What do you think?

asset-registry/src/lib.rs
sander2: Done. However, I'm not sure what use case you envision since there is already the `process_asset` trait function that is able to assign an id. Could you clarify?

enthusiastmartin:
> > I think we should still require some basic fields like name, symbol, decimals. otherwise the js sdk still need specialized code to handle it for different parachains.
> 
> Sure, I added name, symbol, decimals and multilocation.
> 
> > I agree with xlc - that some basic information can be already part of the pallet ( name,symbol, decimals).
> > personally, i would not require `Into<MultiLocation>` for asset metadata, and i would not map the location to asset id when asset is registered. I'd add ability to set/update/remove location explicitly after asset is already registered.
> 
> When I added default fields, I also added the (multi)location. Are you opposed to this? Do you think it should be an optional item? Or should it be removed as a default field altogether?

I might be wrong here - but i think, if you set a location for asset and then xcm config is set to map asset and locations - that will make the asset automatically transferable. 

which is probably fine, that's why you add foreign assets in first place.

but not all assets need location, and not all asset are transferable. For example - in HydraDX's xyk pool - we use asset registry to register 'share asset' for each subpool as another asset in the system. However, this asset does not need location nor it can be transferred somewhere else.

sander2:
> I might be wrong here - but i think, if you set a location for asset and then xcm config is set to map asset and locations - that will make the asset automatically transferable.
> 
> which is probably fine, that's why you add foreign assets in first place.
> 
> but not all assets need location, and not all asset are transferable. For example - in HydraDX's xyk pool - we use asset registry to register 'share asset' for each subpool as another asset in the system. However, this asset does not need location nor it can be transferred somewhere else.

Thank you for the clarification, that makes sense indeed. I changed the location to be a `Option<MultiLocation>`, and also added a `set_asset_location` function. In your usecase you'd set the location in the metadata to `None`, while xcm transferable assets can still be registered in a single step using only `register_asset`. The reason why I think it's good to be able to register xcm-transferrable assets in a single call, is that making multiple calls is problematic if you have to go through governance. You could also batch `register_asset` and `set_asset_location`, but the problem there is that the asset id might not be known in advance in all cases

enthusiastmartin:
What about existential deposit / min balance ? 

I believe this could be one of the required fields too. 

i think that it would be useful if you could also plug asset registry to eg. orml_tokens's config `ExistentialDeposits`. 
that would mean, it needs `GetByKey` implementation. 

Eg. in our asset-registry [here](https://github.com/galacticcouncil/Basilisk-node/blob/afb76f0c80380fc59d802af62c5a50f85b6ba213/pallets/asset-registry/src/lib.rs#L473).

sander2:
> What about existential deposit / min balance ?
> 
> I believe this could be one of the required fields too.
> 
> i think that it would be useful if you could also plug asset registry to eg. orml_tokens's config `ExistentialDeposits`. that would mean, it needs `GetByKey` implementation.
> 
> Eg. in our asset-registry [here](https://github.com/galacticcouncil/Basilisk-node/blob/afb76f0c80380fc59d802af62c5a50f85b6ba213/pallets/asset-registry/src/lib.rs#L473).

Agreed, added.

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+pub trait AssetProcessor<AssetId, Metadata> {
+	fn process_asset(id: Option<AssetId>, asset_metadata: &Metadata) -> Result<(AssetId, Metadata), DispatchError>;
+}
+
+#[derive(scale_info::TypeInfo, Encode, Decode, Clone, Eq, PartialEq, Debug)]
+pub struct AssetMetadata<T: Parameter + Member + TypeInfo> {
+	pub decimals: u32,
+	pub name: Vec<u8>,
+	pub symbol: Vec<u8>,
+	pub location: Option<MultiLocation>,
```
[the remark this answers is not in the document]
sander2: Good question, I'm not sure.. I think `MultiLocation` is easiest to deal with in the parachains, but for off-chain interactions (e.g. reading from js) a versionedLocation might make sense. Anyone else has any thoughts on this?

asset-registry/src/lib.rs
xlc: Our asset registry supports not only foreign assets, but also native assets. So that asset_id could be an existing one instead of new id.

asset-registry/src/lib.rs
xlc: The default SequentialId can still read the last id from the pallet?

asset-registry/src/lib.rs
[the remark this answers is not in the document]
xlc: Currently `MultiLocation` is referring to the latest version, which will change in future and causing expected issue.
So instead of using the latest, it should use v2 (and move to v3 later if needed) so ensure it doesn't change unexpectedly.

`VersionedMultiLocation` will also avoid this issue here

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+	/// The total issuance of a token type.
+	#[pallet::storage]
+	#[pallet::getter(fn get_metadata)]
+	pub type Metadata<T: Config> =
+		StorageMap<_, Twox64Concat, T::AssetId, AssetMetadata<T::Balance, T::CustomMetadata>, OptionQuery>;
+
+	/// The total issuance of a token type.
+	#[pallet::storage]
+	#[pallet::getter(fn get_asset_id)]
+	pub type MultiLocationLookup<T: Config> = StorageMap<_, Twox64Concat, MultiLocation, T::AssetId, OptionQuery>;
```
xlc, line 89: We can't use xcm::latest::MultiLocation as key format here because it will change.
We can't use VersionedMultiLocation here because it will result inconsistency.
What polkadot xcm pallet doing is using VersionedMultiLocation but always force it to be the latest version and using runtime migration to keep it up to date. But that logic is super complicated.
I will say just use v2::MultiLocation for now and migrate to v3 if needed.

sander2: Done

sander2:
I cleaned the code up a bit and added some tests, including some that show that it can be used in xcm transfers. From my PoV, this PR is ready to be reviewed.

There is one thing that I still need address: the default weights. For the other pallets, the default weights are generated with `acala/target/release/acala benchmark`. Since this pallet is not included in that repo, should I add benchmark code in the crate itself?

One more note on the `ProcessAsset` trait. By request of @xlc it's returning `(AssetId, Metadata)`, meaning that it can modify the supplied metadata. I'm not certain what the use case of this is, but running the same function again when _modifying_ metadata is problematic, so I only call it in the `register_asset` function and not in `update_asset`.

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,243 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::large_enum_variant)]
```
xlc, line 2: please fix the warning instead of suppress it. in this case you should box VersionedMultiLocation

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+	#[pallet::getter(fn get_asset_id)]
+	pub type MultiLocationLookup<T: Config> = StorageMap<_, Twox64Concat, MultiLocation, T::AssetId, OptionQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn last_asset_id)]
+	pub(crate) type LastAssetId<T: Config> = StorageValue<_, T::AssetId, OptionQuery>;
+
+	#[pallet::genesis_config]
+	pub struct GenesisConfig<T: Config> {
+		_phantom: PhantomData<T>,
```
xlc, line 115: should have a way to register assets in genesis, but it can happen in later PR

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+		fn build(&self) {}
+	}
+
+	#[pallet::pallet]
+	#[pallet::generate_store(pub(super) trait Store)]
+	#[pallet::without_storage_info]
```
xlc, line 134: should use bonded types, but that can happen in later PR

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+
+		/// The type used as a unique asset id,
+		type AssetId: Parameter + Member + TypeInfo;
+
+		/// The origin that is allowed to manipulate metadata.
+		type AuthorityOrigin: EnsureOrigin<<Self as frame_system::Config>::Origin>;
+
+		/// A filter ran upon metadata registration that assigns an is and
+		/// potentially modifies the supplied metadata.
+		type ProcessAsset: AssetProcessor<Self::AssetId, AssetMetadata<Self::Balance, Self::CustomMetadata>>;
```
xlc, line 53: maybe call it `OnRegisterAsset` or something similar

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+		let asset_id = MultiLocationLookup::<T>::get(location)?;
+		Metadata::<T>::get(asset_id)
+	}
+
+	fn insert_metadata(
+		asset_id: &T::AssetId,
+		metadata: &AssetMetadata<T::Balance, T::CustomMetadata>,
+	) -> DispatchResult {
+		// if the metadata contains a location, set the MultiLocationLookup
+		if let Some(location) = metadata.location.clone() {
```
xlc, line 229: I think this means when update asset metadata without changing location, the location is still going to be removed & inserted.

Not end of world as we shouldn't update asset metadata every block.

asset-registry/src/lib.rs
sander2: Done, but I think this warning was maybe a false positive, since the other items contain `AssetMetadata`, which has a `Option<VersionedMultiLocation>` field, so as far as I can tell the sizes should have been similar to each other

asset-registry/src/impls.rs
```diff
@@ -0,0 +1,163 @@
...
+pub struct SequentialId<AssetId, Metadata, T>(PhantomData<(AssetId, Metadata, T)>);
+
+impl<AssetId: AtLeast32BitUnsigned, Metadata, T> AssetProcessor<AssetId, Metadata>
+	for SequentialId<AssetId, Metadata, T>
+where
+	AssetId: AtLeast32BitUnsigned + Parameter + Member + TypeInfo,
+	T: Config<AssetId = AssetId>,
+{
+	fn process_asset(id: Option<AssetId>, asset_metadata: Metadata) -> Result<(AssetId, Metadata), DispatchError> {
+		let next_id = match LastAssetId::<T>::get() {
```
xlc, line 25: `LastAssetId::<T>::get().unwrap_or_default().saturating_add(One::one())`

xlc (commented):
not 100% sure if we want all the XCM weight related code in this pallet. I guess we don't really want another crate just for that so it is probably fine.

asset-registry/src/lib.rs
xlc: It could be false negative. It is known issue that big enum of Call / Event can cause trouble.

asset-registry/src/lib.rs
sander2: You are right. However, since as you say, this shouldn't be called very frequently. We could optimize in the future but for now I prioritized code simplicity over performance

asset-registry/src/lib.rs
sander2: Hmm yea I actually just ran into a problem when trying to implement that: it requires the `AssetMetadata` to implement Serialize & Deserialize, which `VersionedMultiLocation` does not implement.. We'd have to write a custom implementation, which is certainly possible, but I'd leave it for a future PR indeed

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+			asset_id: T::AssetId,
+			metadata: AssetMetadata<T::Balance, T::CustomMetadata>,
+		},
+		SetLocation {
+			asset_id: T::AssetId,
+			location: Box<VersionedMultiLocation>,
+		},
+	}
+
+	/// The total issuance of a token type.
```
xlc, line 98: bad comment

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn get_metadata)]
+	pub type Metadata<T: Config> =
+		StorageMap<_, Twox64Concat, T::AssetId, AssetMetadata<T::Balance, T::CustomMetadata>, OptionQuery>;
+
+	/// The total issuance of a token type.
```
xlc, line 104: bad comment

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn get_asset_id)]
```
xlc, line 106: the getter name doesn't align with the storage name
also maybe we want to call it `LocationToAssetId`?

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+		},
+		SetLocation {
+			asset_id: T::AssetId,
+			location: Box<VersionedMultiLocation>,
+		},
+	}
+
+	/// The total issuance of a token type.
+	#[pallet::storage]
+	#[pallet::getter(fn get_metadata)]
```
xlc, line 100:
```suggestion
	#[pallet::getter(fn metadata)]
```

no `get_` in getter
https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

xlc:
other than nits it is good for a V1

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+		if let Some(location) = metadata.location.clone() {
+			let location: MultiLocation = location.try_into().map_err(|()| Error::<T>::BadVersion)?;
+			ensure!(
+				!MultiLocationLookup::<T>::contains_key(&location),
+				Error::<T>::ConflictingLocation
+			);
+			MultiLocationLookup::<T>::insert(location, asset_id);
+		}
+
+		Metadata::<T>::insert(&asset_id, &metadata);
```
zqhxuyuan, line 238: We could add a judegment before insert operation, if asset_id exist, throw AssetIdExist Error.

zqhxuyuan: I'm not sure `try_mutate` or `insert` which one is best:
```
Metadata::<T>::try_mutate(
	&asset_id,
	|maybe_asset_metadatas| -> DispatchResult {
		ensure!(maybe_asset_metadatas.is_none(), Error::<T>::AssetIdExisted);

		*maybe_asset_metadatas = Some(metadata.clone());
		Ok(())
	},
)?;

Ok(())
```

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use frame_support::pallet_prelude::*;
+use frame_support::traits::EnsureOrigin;
+use frame_support::transactional;
```
zqhxuyuan, line 5: use tight format: `use frame_support::{....}`, also other one in this file and other files needs to tidy-up.

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+		Ok(())
+	}
+
+	pub fn fetch_metadata_by_location(
+		location: &MultiLocation,
+	) -> Option<AssetMetadata<T::Balance, T::CustomMetadata>> {
+		let asset_id = MultiLocationLookup::<T>::get(location)?;
+		Metadata::<T>::get(asset_id)
+	}
```
zqhxuyuan, line 222: add another two helper method: `get_assetid(multilocation)` and `get_multilocation(asset_id)` which is helpful to `AssetId` <-> `MultiLocation` conversion.

traits/src/asset_registry.rs
```diff
@@ -0,0 +1,14 @@
...
+pub trait WeightToFeeConverter {
+	fn convert_weight_to_fee(location: &MultiLocation, weight: Weight) -> Option<u128>;
+}
+
+pub trait FixedConversionRateProvider {
+	fn get_fee_per_second(location: &MultiLocation) -> Option<u128>;
+}
+
+pub trait AssetProcessor<AssetId, Metadata> {
+	fn process_asset(id: Option<AssetId>, asset_metadata: Metadata) -> Result<(AssetId, Metadata), DispatchError>;
```
zqhxuyuan, line 13: rename this method to `pre_register(..)` and add another method named: `post_register()`
```rust
fn pre_register(id: Option<AssetId>, asset_metadata: Metadata) -> Result<(AssetId, Metadata), DispatchError>;
fn post_register(id: Option<AssetId>, asset_metadata: Metadata) -> Result<(AssetId, Metadata), DispatchError>;
```

As customized parachains may want to do additional behaviour after asset registed.

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+		#[pallet::weight(T::WeightInfo::register_asset())]
+		#[transactional]
+		pub fn register_asset(
+			origin: OriginFor<T>,
+			metadata: AssetMetadata<T::Balance, T::CustomMetadata>,
+			asset_id: Option<T::AssetId>,
+		) -> DispatchResult {
+			let _ = T::AuthorityOrigin::ensure_origin(origin)?;
+
+			let (asset_id, metadata) = T::OnRegisterAsset::process_asset(asset_id, metadata)?;
```
zqhxuyuan, line 151: I think it'll be good if move `process_asset` processing part into `insert_metdata()`. for parachains who want to use their own dispatch call and intergrate orml, they can use `orml_asset_registry::Pallet::insert_metadata()`

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+			Self::insert_metadata(&asset_id, &metadata)?;
```
zqhxuyuan, line 153: maybe rename `insert_metadata` to `do_register_asset`, and rename `update_metadata` to `do_update_asset`?

asset-registry/src/lib.rs
sander2: That's a bit problematic, since `insert_metadata` is also used from within the `update_asset` function. Should we also call that in `update_asset` and `set_location`?

asset-registry/src/lib.rs
sander2: I added the check in `register_asset` - I named the error `ConflictingAssetId`, to be consistent with the existing `ConflictingLocation` error.

I also updated the code to use try_mutate where possible. Instead of `AssetIdNotExisted` I'm reusing the existing `AssetNotFound`

asset-registry/src/lib.rs
sander2: `get_assetid` was already available through the getter generated by the storage item. I implemented `get_multilocation` but named it `get_multilocation` because of [this comment](https://github.com/open-web3-stack/open-runtime-module-library/pull/728#discussion_r867281890).

traits/src/asset_registry.rs
sander2: Done (except the type is `fn post_register(_id: AssetId, _asset_metadata: Metadata) -> Result<(), DispatchError>`).

Currently they are only called in `register_asset`, not in `update_asset`. Not sure if we should have hooks there as well..

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+			ensure!(maybe_metadata.is_none(), Error::<T>::ConflictingAssetId);
+
+			*maybe_metadata = Some(metadata.clone());
+
+			Ok(())
+		})?;
+
+		if let Some(ref location) = metadata.location {
+			Self::do_insert_location(asset_id.clone(), location.clone())?;
+		}
```
zqhxuyuan, line 210: move `do_insert_location` to inner of `Metadata::<T>::try_mutate()`

traits/src/asset_registry.rs
zqhxuyuan: `update_asset` don't need hook for now. but if someone have those customize, it can be add then.

asset-registry/src/mock/para.rs
```diff
@@ -0,0 +1,337 @@
...
+pub type AssetRegistryWeightTrader =
+	(AssetRegistryTrader<FixedRateAssetRegistryTrader<MyFixedConversionRateProvider>, ToTreasury>,);
+
+pub struct MyFixedConversionRateProvider;
+impl FixedConversionRateProvider for MyFixedConversionRateProvider {
+	fn get_fee_per_second(location: &MultiLocation) -> Option<u128> {
+		let metadata = AssetRegistry::fetch_metadata_by_location(location)?;
+		Some(metadata.additional.fee_per_second)
+	}
+}
```
zqhxuyuan, line 182: Can this move to impl.rs, I think this is a common implementation.

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(T::WeightInfo::register_asset())]
+		#[transactional]
+		pub fn register_asset(
+			origin: OriginFor<T>,
+			metadata: AssetMetadata<T::Balance, T::CustomMetadata>,
+			asset_id: Option<T::AssetId>,
+		) -> DispatchResult {
+			let _ = T::AuthorityOrigin::ensure_origin(origin)?;
```
shaunxw, line 146:
```suggestion
			T::AuthorityOrigin::ensure_origin(origin)?;
```

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,273 @@
...
+					if let Some(ref location) = old_metadata.location {
+						let location: MultiLocation =
+							location.clone().try_into().map_err(|()| Error::<T>::BadVersion)?;
+						LocationToAssetId::<T>::remove(location);
+					}
+
+					// insert new location
+					if let Some(ref location) = metadata.location {
+						Self::do_insert_location(asset_id.clone(), location.clone())?;
+					}
```
shaunxw, line 237: Need to check if location changed before remove/insert.

asset-registry/src/tests.rs
```diff
@@ -0,0 +1,377 @@
+#![cfg(test)]
```
shaunxw, line 1: Unit tests should cover all pallet errors, the following is not covered yet:
- AssetNotFound
- BadVersion
- InvalidAssetId

shaunxw (commented):
I suggest to change `update_asset` dispatchable call signature to

```rust
pub fn update_asset(
			origin: OriginFor<T>,
			asset_id: T::AssetId,
			decimals: Option<u32>,
			name: Option<Vec<u8>>,
			// ... rest fields wrapped in `Option`
		) -> DispatchResult
```

The impl checks every field and update it if `Some`. Note `location` needs to be `Option<Option<VersionedMultiLocation>>`.

The rationale behind it is that in most cases, only one field will be updated, but the current call requires to copy all existing unchanged ones, which is unnecessary and prune to mistakes. Additionally, `set_asset_location` call can be merged into this one.

asset-registry/src/impls.rs
```diff
@@ -0,0 +1,162 @@
...
+	fn drop(&mut self) {
+		if let Some(ref bought) = self.bought_weight {
+			R::take_revenue((AssetId::Concrete(bought.asset_location.clone()), bought.amount).into());
+		}
+	}
+}
+
+// Return Existential deposit of an asset. Implementing this trait allows the
+// pallet to be used in the tokens::ExistentialDeposits config item
+impl<T: Config> GetByKey<T::AssetId, T::Balance> for Pallet<T> {
```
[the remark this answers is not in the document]
sander2: Done, although it needs to be generic over the Config, so usage is now `AssetRegistryExistentialDeposits::<Runtime>::get(asset_id)`

asset-registry/src/impls.rs
```diff
@@ -0,0 +1,162 @@
...
+
+				let (existing_weight, existing_fee) = match self.bought_weight {
+					Some(ref x) => (x.weight, x.amount),
+					None => (0, 0),
+				};
+
+				let new_weight = existing_weight.saturating_add(weight);
+
+				if let Some(amount) = W::convert_weight_to_fee(location, new_weight) {
+					let fee_increase = amount.saturating_sub(existing_fee);
```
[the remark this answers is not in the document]
sander2: > I might miss something, but not sure why adding weight to existing to get new_weight first, then convert to fee and subtract. Simply converting weight to fee should have the same result no?

It's equivalent only when `convert_weight_to_fee` is a linear function, which it is not guaranteed to be. Though maybe I'm overcomplicating things - I don't expect to see non-linear functions being used in practice..

> And the embedded if/if let checks could be flattened for readability, such as:
> ```if let Some(amount) = W::convert_weight_to_fee(location, new_weight) {```
> -->
> ```let amount = W::convert_weight_to_fee(location, new_weight).ok_or(XcmError::TooExpensive)?;```

That would stop the iteration over the assets in `payment.fungible`. The same applies for the last refactoring you suggest

asset-registry/src/tests.rs
sander2: It's not so easy to test the BadVersion error though - that's the error you get when converting locations fails. I am not even sure it is possible.. It's also not being tested in the rest of the project

sander2:
I think I've addressed all comments - PR should be ready for final review pass & merging

asset-registry/src/impls.rs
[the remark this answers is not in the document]
shaunxw: I don't see non-linear functions being used either. AFAIK ppl are all using fixed rate conversions, specifically the fixed rate weight traders implemented in Polkadot repo.

IMO we can simply aim for the same impl to provide a fixed rate trader. It's easier to use and maintain.

asset-registry/README.md
```diff
@@ -0,0 +1,5 @@
+# Asset Registry Module
+
+## Overview
+
+This module
```
shaunxw, line 5: Missed this one in my previous review. Could you please finished the readme?

asset-registry/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+license = "Apache-2.0"
+version = "0.4.1-dev"
+authors = ["Interlay Ltd, etc"]
+edition = "2021"
+
+[dependencies]
+scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
+serde = { version = "1.0.136", optional = true }
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["max-encoded-len"] }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.18", default-features = false }
```
shaunxw, line 14: Please also update deps to `v0.9.20`.

shaunxw (commented):
LGTM. A few more trivial issues and I think we are good to go.

sander2:
Fixed the last comments and rebased (and squashed) the branch.

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,312 @@
...
+					if let Some(additional) = additional {
+						(*metadata).additional = additional;
+					}
+
+					Ok(())
+				}
+			}
+		})?;
+
+		// Self::deposit_event(Event::<T>::UpdatedAsset { asset_id, metadata });
```
shaunxw, line 282: Will need to deposit event, probably in `Some(ref mut metadata) => {` branch.

shaunxw (approved):

asset-registry/src/lib.rs
sander2: Oops, fixed

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+		Self::do_register_asset_without_asset_processor(metadata.clone(), asset_id.clone())?;
+
+		T::AssetProcessor::post_register(asset_id, metadata)?;
+
+		Ok(())
+	}
+
+	// This is useful in tests but it might also come in useful to users of this
+	// pallet.
+	/// Like do_register_asset, but without calling pre_register and
```
zqhxuyuan, line 202: use `///` above

asset-registry/src/impls.rs
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+/// A default implementation for WeightToFeeConverter that takes a fixed
+/// conversion rate.
+pub struct FixedRateAssetRegistryTrader<P: FixedConversionRateProvider>(PhantomData<P>);
+impl<P: FixedConversionRateProvider> WeightToFeeConverter for FixedRateAssetRegistryTrader<P> {
+	fn convert_weight_to_fee(location: &MultiLocation, weight: Weight) -> Option<u128> {
+		let fee_per_second = P::get_fee_per_second(location)?;
+		let amount = fee_per_second.saturating_mul(weight as u128) / (WEIGHT_PER_SECOND as u128);
```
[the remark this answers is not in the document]
sander2: I didn't think that was necessary since WEIGHT_PER_SECOND is a constant, but I made the change you requested regardless

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+		#[pallet::weight(T::WeightInfo::update_asset())]
+		#[transactional]
+		pub fn update_asset(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+			decimals: Option<u32>,
+			name: Option<Vec<u8>>,
+			symbol: Option<Vec<u8>>,
+			existential_deposit: Option<T::Balance>,
+			location: Option<Option<VersionedMultiLocation>>,
```
zqhxuyuan, line 165: we can simplify this parameter as `Option<VersionedMultiLocation>`.

sander2: We can't - the outer Option determines whether or not to update that parameter, whereas the inner one is there because the user might want to unset the location, i.e. to make `metadata.location = None`

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+		decimals: Option<u32>,
+		name: Option<Vec<u8>>,
+		symbol: Option<Vec<u8>>,
+		existential_deposit: Option<T::Balance>,
+		location: Option<Option<VersionedMultiLocation>>,
+		additional: Option<T::CustomMetadata>,
+	) -> DispatchResult {
+		Metadata::<T>::try_mutate(&asset_id, |maybe_metadata| -> DispatchResult {
+			match maybe_metadata {
+				None => Err(Error::<T>::AssetNotFound.into()),
```
zqhxuyuan, line 237: replace with below, and remove match ase. also replace `(*metadata)` with `metadata`
```rust
let metadata = maybe_metadata.as_mut().ok_or(Error::<T>::AssetNotFound)?;
```

asset-registry/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+								let location: MultiLocation =
+									location.clone().try_into().map_err(|()| Error::<T>::BadVersion)?;
+								LocationToAssetId::<T>::remove(location);
+							}
+
+							// insert new location
+							if let Some(ref new_location) = location {
+								Self::do_insert_location(asset_id.clone(), new_location.clone())?;
+							}
+						}
```
zqhxuyuan, line 269: would be better to use `LocationToAssetId::try_mutate(location, |maybe_asset_id| {....})`

sander2: We are dealing with 2 different keys though - we have `old_location` and `new_location`, so it can't be combined into a single `try_mutate`. Setting `new_location` does use `try_mutate` inside `do_insert_location`, but for `old_location` the remove is more straightforward

PR #732, opened 2022-05-03
[feat] extend NamedMultiReservable

### LINKED ISSUE #731: Improve MultiNamedReservableCurrency
Followup of #730

Add new methods similar to

https://github.com/paritytech/substrate/blob/c4693ebfa67f4aafbf66daa7b70a8b88f17bddc7/frame/support/src/traits/tokens/currency/reservable.rs#L180-L237

stanly-johnson:
Changelog:
- Adds ensure_reserved_named, unreserve_all_named, slash_all_reserved_named, repatriate_all_reserved_named functions to NamedMultiReservable Currency.

Resolves #731

### DIFF
--- a/tokens/src/tests.rs
+++ b/tokens/src/tests.rs
@@ -1773,6 +1773,94 @@ fn slashed_reserved_named_works() {
 		});
 }
 
+#[test]
+fn named_multi_reservable_ensure_named_reserved_works() {
+	ExtBuilder::default()
+		.balances(vec![(ALICE, DOT, 100)])
+		.build()
+		.execute_with(|| {
+			assert_ok!(Tokens::reserve_named(&RID_1, DOT, &ALICE, 50));
+			assert_eq!(Tokens::free_balance(DOT, &ALICE), 50);
+			assert_eq!(Tokens::reserved_balance_named(&RID_1, DOT, &ALICE), 50);
+			assert_eq!(Tokens::total_issuance(DOT), 100);
+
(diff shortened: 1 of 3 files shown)


tokens/src/tests.rs
```diff
@@ -1773,6 +1773,94 @@ fn slashed_reserved_named_works() {
 		});
 }
 
+#[test]
```
zqhxuyuan, line 1776: this is a very huge test file, should we split into multi test file under tests folder? @xlc

xlc: #733 doesn't need to be a blocker for this

traits/src/currency.rs
```diff
@@ -288,6 +288,77 @@ pub trait NamedMultiReservableCurrency<AccountId>: MultiReservableCurrency<Accou
...
+		currency_id: Self::CurrencyId,
+		who: &AccountId,
+	) -> Self::Balance {
+		let value = Self::reserved_balance_named(id, currency_id, who);
+		Self::unreserve_named(id, currency_id, who, value);
+		value
+	}
+
+	/// Slash all the reserved balance, returning the negative imbalance
+	/// created.
```
zqhxuyuan, line 332: current implementation of `slash_reserved_named` only return one value, not returning nagative imbalance.
comment here can change to: returning the amount that was unable to be slashed.

or we could align at substrate: https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/currency/reservable.rs#L124 which return tuple in `slash_reserved_named`, but I think this can be a later issue.

xlc (approved):

Issue #742, opened 2022-05-17
NamedMultiReservableCurrency trait bound for orml_currencies `MultiCurrency`


Chralt98:
Hey there,

after implementing the `NamedMultiReservalbeCurrency`, is it straight forward to [add it to the trait bound of `MultiCurrency`](https://github.com/open-web3-stack/open-runtime-module-library/blob/39057dae9c277e34a8e1c61abf2148158199e687/currencies/src/lib.rs#L84-L87) in the orml_currencies pallet?

This would be cool. I don't really know what it means for other dependencies yet.

xlc:
Why?

Chralt98:
This would add the functionality of adding an `ReserveIdentifier` id to the MultiCurrency  in `orml_currencies`.

PR #743, opened 2022-05-17
Implement NamedMultiReservableCurrency and NamedBasicReservableCurrency in orml_currencies

Chralt98:
Hey, this is related to #742. 

It's the first try and it's not ready yet.

### DIFF
--- a/currencies/src/lib.rs
+++ b/currencies/src/lib.rs
@@ -372,6 +372,8 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
 }
 
 impl<T: Config> NamedMultiReservableCurrency<T::AccountId> for Pallet<T> {
+	type ReserveIdentifier = [u8; 8];
+
 	fn slash_reserved_named(
 		id: &Self::ReserveIdentifier,
 		currency_id: Self::CurrencyId,
@@ -567,12 +569,9 @@ where
 	T: Config,
 	GetCurrencyId: Get<CurrencyIdOf<T>>,
 {
-	fn slash_reserved_named(
-		id: &Self::ReserveIdentifier,
-		currency_id: Self::CurrencyId,
-		who: &AccountId,
-		value: Self::Balance,
-	) -> Self::Balance {
+	type ReserveIdentifier = [u8; 8];
+
+	fn slash_reserved_named(id: &Self::ReserveIdentifier, who: &T::AccountId, value: Self::Balance) -> Self::Balance {
 		<Pallet<T> as NamedMultiReservableCurrency<T::AccountId>>::slash_reserved_named(
 			id,
 			GetCurrencyId::get(),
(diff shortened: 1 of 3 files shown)


xlc:
Ok. Simply add the trait bound makes no sense (that's what I read from the issue) but actually implement `NamedMultiReservableCurrency` is useful.

Chralt98:
@xlc I don't get why the cargo fmt isn't correct. I formatted it multiple times.

Chralt98:
I let you finish the work. I think the most is done.

currencies/src/lib.rs
```diff
@@ -84,11 +85,13 @@ pub mod module {
 		type MultiCurrency: TransferAll<Self::AccountId>
 			+ MultiCurrencyExtended<Self::AccountId>
 			+ MultiLockableCurrency<Self::AccountId>
-			+ MultiReservableCurrency<Self::AccountId>;
+			+ MultiReservableCurrency<Self::AccountId>
+			+ NamedMultiReservableCurrency<Self::AccountId, ReserveIdentifier = [u8; 8]>;
```
zqhxuyuan, line 89: It's better to define a ReserveIdentifier type config here.

xlc: There isn't a real need to define it. Just leave it out and use `ReserveIdentifierOf` when you need it.

`type ReserveIdentifierOf<T> = <<T as Config>::MultiCurrency as NamedMultiReservableCurrency <<T as frame_system::Config>::AccountId>>:: ReserveIdentifier;`

Chralt98:
I don't really know if [this](https://github.com/open-web3-stack/open-runtime-module-library/blob/c76ec675ea7f3334a0dffbb4d195f49147c34a4a/currencies/src/lib.rs#L779) might be a problem in the future. I mean we got a `ReserveIdentifier` for `orml_tokens` AND another `ReserveIdentifier` for `pallet_balances`.

PS: I hate this cargo fmt problem here in this repo...

xlc:
Update your cargo to a recent version should resolve the fmt issue. This is one used by CI
https://github.com/open-web3-stack/open-runtime-module-library/blob/39057dae9c277e34a8e1c61abf2148158199e687/.github/workflows/test.yml#L36

xlc:
Multiple `ReserveIdentifier` is not an issue. You can require them to be the same one if you need. Similar to how the `Balance` type for pallet-balances and orml-tokens are handled.

Chralt98:
Thanks @xlc ! If there is anything to correct in my pull request, I am open to correct it after a review.

currencies/src/lib.rs
```diff
@@ -619,7 +738,7 @@ where
 impl<T, AccountId, Currency, Amount, Moment> BasicReservableCurrency<AccountId>
 	for BasicCurrencyAdapter<T, Currency, Amount, Moment>
 where
-	Currency: PalletReservableCurrency<AccountId>,
+	Currency: PalletNamedReservableCurrency<AccountId>,
```
xlc, line 741: why change this?

Chralt98: Sorry, this was my bad.

currencies/src/lib.rs
```diff
@@ -368,6 +375,75 @@ impl<T: Config> MultiReservableCurrency<T::AccountId> for Pallet<T> {
 	}
 }
 
+impl<T: Config> NamedMultiReservableCurrency<T::AccountId> for Pallet<T> {
+	type ReserveIdentifier = ReserveIdentifierOf<T>;
```
xlc, line 379:
```suggestion
	type ReserveIdentifier = ReserveIdentifierOf<T>;

```

zqhxuyuan:
LGTM except missing testcase, can you do this here or maybe in another PR? @Chralt98

Chralt98:
@zqhxuyuan This depends on, what would be faster for a merge (Zeitgeist requires this functionality as fast as possible). I would propose, that it should get another pull request. But please verify the code before merge again :-D. I don't like to impact many chains to loose money. I would try to add test cases later in the day in another pull request. But I would be happy, if you could spend time on this too.

zqhxuyuan (approved):
LGTM

zqhxuyuan:
It's better to add `can_reserve` method too, you can do this on later PR. @Chralt98

Chralt98:
@zqhxuyuan You mean `can_reserve_named` for `NamedBasicReservableCurrency` and for `NamedMultiReservableCurrency` then? But this is not implemented for `orml_tokens`. Or what do you mean?

zqhxuyuan:
ok, my bad. the `can_reseve` is not needed for named implementation.

Issue #751, opened 2022-05-28
Support `TransferReserveAsset` to transfer asset


zqhxuyuan:
For case like ERC20 cross chain transfer: https://github.com/AcalaNetwork/Acala/issues/2167, evm don't support withdraw and deposit, but support transfer. while `TransferReserveAsset` can be replace withdraw and deposit, like the usage in relaychain `reserve_transfer_asset`.

PR #752, opened 2022-05-28
support `TransferReserveAsset` to transfer

zqhxuyuan:
relates to: https://github.com/open-web3-stack/open-runtime-module-library/issues/751

### DIFF
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -51,6 +51,8 @@ mod tests;
 enum TransferKind {
 	/// Transfer self reserve asset.
 	SelfReserveAsset,
+	/// Use `transfer` to transfer self reserve asset.
+	TransferReserve,
 	/// To reserve location.
 	ToReserve,
 	/// To non-reserve location.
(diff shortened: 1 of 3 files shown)


xlc (commented):
Maybe it is actually better to always use transfer reserve instead withdraw & deposit?
I don't see too much reason to use the old transfer anymore, but I guess we need to keep it to avoid unexpected breaking change to other parachains.

xlc (approved):

xtokens/src/lib.rs
```diff
@@ -761,13 +823,17 @@ pub mod module {
 		fn transfer_kind(
 			reserve: Option<MultiLocation>,
 			dest: &MultiLocation,
+			transfer_kind: Option<TransferKind>,
```
shaunxw, line 826: Transfer kind should be determined by `reserve` and `dest` location. This kind of override is bug prone. Edge cases should be handled outside this function.

shaunxw:
The `SelfReserveAsset` and `ReserveTransfer` impl looks redundant to me.

Can we just update `transfer_self_reserve_asset` impl to use `TransferReserveAsset` instruction? https://github.com/open-web3-stack/open-runtime-module-library/blob/070457de18d26d2daed6abdfa57d8a951a525605/xtokens/src/lib.rs#L638-L657

shaunxw (approved):
LGTM

PR #753, opened 2022-05-29
Distinguish when `MinXcmFee` is not registered and `FeeNotEnough`

ghzlatarev:
Signed-off-by: Georgi Zlatarev <georgi.zlatarev@manta.network>

A few small changes to improve debugging and readability:
* Added a specific error to specify `MinXcmFee` is not registered, to distinguish between this case and `FeeNotEnough`
* Refactored `MinXcmFee`  return an Option, so implementers don't have to use u128::Max for unregistered locations. 
* Adjusted existing test cases

### DIFF
--- a/traits/src/get_by_key.rs
+++ b/traits/src/get_by_key.rs
@@ -1,5 +1,3 @@
-use xcm::v1::MultiLocation;
-
 /// A trait for querying a value by a key.
 pub trait GetByKey<Key, Value> {
 	/// Return the value.
@@ -56,12 +54,4 @@ mod tests {
 		assert_eq!(Test::get(&2), 2);
 		assert_eq!(Test::get(&3), 2);
 	}
-}
-
-// Default implementation for xTokens::MinXcmFee
-pub struct DisabledParachainFee;
-impl GetByKey<MultiLocation, Option<u128>> for DisabledParachainFee {
-    fn get(_key: &MultiLocation) -> Option<u128> {
-        None
-    }
 }
--- a/traits/src/get_by_key.rs
+++ b/traits/src/get_by_key.rs
@@ -1,3 +1,5 @@
+use xcm::v1::MultiLocation;
+
 /// A trait for querying a value by a key.
 pub trait GetByKey<Key, Value> {
 	/// Return the value.
@@ -54,4 +56,12 @@ mod tests {
 		assert_eq!(Test::get(&2), 2);
 		assert_eq!(Test::get(&3), 2);
 	}
+}
+
+// Default implementation for xTokens::MinXcmFee
+pub struct DisabledParachainFee;
+impl GetByKey<MultiLocation, Option<u128>> for DisabledParachainFee {
+    fn get(_key: &MultiLocation) -> Option<u128> {
+        None
+    }
 }
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -174,7 +174,7 @@ pub mod module {
 		/// Not supported MultiLocation
 		NotSupportedMultiLocation,
 		/// MinXcmFee not registered for certain reserve location
-		MinXcmFeeNotRegistered,
+		MinXcmFeeNotDefined,
 	}
 
 	#[pallet::hooks]
@@ -540,7 +540,7 @@ pub mod module {
 				ensure!(non_fee_reserve == dest.chain_part(), Error::<T>::InvalidAsset);
 
 				let reserve_location = non_fee_reserve.clone().ok_or(Error::<T>::AssetHasNoReserve)?;
-				let min_xcm_fee = T::MinXcmFee::get(&reserve_location).ok_or(Error::<T>::MinXcmFeeNotRegistered)?;
+				let min_xcm_fee = T::MinXcmFee::get(&reserve_location).ok_or(Error::<T>::MinXcmFeeNotDefined)?;
 
 				// min xcm fee should less than user fee
 				let fee_to_dest: MultiAsset = (fee.id.clone(), min_xcm_fee).into();
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -89,7 +89,7 @@ pub mod module {
 		type SelfLocation: Get<MultiLocation>;
 
 		/// Minimum xcm execution fee paid on destination chain.
-		type MinXcmFee: GetByKey<MultiLocation, u128>;
+		type MinXcmFee: GetByKey<MultiLocation, Option<u128>>;
 
 		/// XCM executor.
 		type XcmExecutor: ExecuteXcm<Self::Call>;
@@ -173,6 +173,8 @@ pub mod module {
 		FeeNotEnough,
 		/// Not supported MultiLocation
 		NotSupportedMultiLocation,
+		/// MinXcmFee not registered for certain reserve location
+		MinXcmFeeNotRegistered,
 	}
 
 	#[pallet::hooks]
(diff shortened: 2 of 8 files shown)


xtokens/src/lib.rs
```diff
@@ -89,7 +89,7 @@ pub mod module {
 		type SelfLocation: Get<MultiLocation>;
 
 		/// Minimum xcm execution fee paid on destination chain.
-		type MinXcmFee: GetByKey<MultiLocation, u128>;
+		type MinXcmFee: GetByKey<MultiLocation, Option<u128>>;
```
zqhxuyuan, line 92: can we implements a default disabled struct which do not support it, then runtime can just use `DisabledParachainFee`?
```rust
pub struct DisabledParachainFee;
impl GetByKey<MultiLocation, Option<u128>> for DisabledParachainFee {
    fn get(k: &MultiLocation) -> Option<u128> {
        None
    }
}
```

ghzlatarev: Yeah sounds good.

ghzlatarev: I was thinking to put it in either traits or xcm-support, since xtokens itself does not have any other default impl. 
Chose traits so you don't have to import another crate just for this.

traits/src/get_by_key.rs
```diff
@@ -55,3 +57,11 @@ mod tests {
...
 	}
 }
+
+// Default implementation for xTokens::MinXcmFee
+pub struct DisabledParachainFee;
+impl GetByKey<MultiLocation, Option<u128>> for DisabledParachainFee {
+    fn get(_key: &MultiLocation) -> Option<u128> {
+        None
+    }
+}
```
xlc, line 67:
```suggestion
}

```

xlc, line 67:
```suggestion
}

```

xlc, line 67:
```suggestion
}

```

xlc:
please use `cargo fmt` to format

traits/src/get_by_key.rs
```diff
@@ -55,3 +57,11 @@ mod tests {
 		assert_eq!(Test::get(&3), 2);
 	}
 }
+
+// Default implementation for xTokens::MinXcmFee
+pub struct DisabledParachainFee;
```
shaunxw, line 62: This should be defined in `orml-xcm-support` pallet.

ghzlatarev:
> please use `cargo fmt` to format



> please use `cargo fmt` to format

should be clean now.

xlc:
Test is failing

ghzlatarev:
Btw I have a related test case that I'm not sure if it's an issue in my setup or the xtokens code. The case is the following:
Alice has 100 of ParaB's native token on ParaA and wants to send it back to ParaB, but wants the fee to be paid with 50 of the relay chain asset. The `MinXcmFee` for ParaB's native token reserve is set to 40 on ParaA.
On the sender side, the `fee > min_xcm_fee` so we don't hit `FeeNotEnough` error and both xcm messages get dispatched to ParaB.
Then the first message arrives on ParaB (via the relay chain), and eventually deposits `assets - min_xcm_fee` ( 50 - 40 = 10 with free execution) of the relay chain asset into ParaA's sovereign account on ParaB. 
Then the second message arrives on ParaB and tries to withdraw the rest of the fee (40, which is the min xcm fee) from ParaA's sovereign account on ParaB, but it only has 10, so it fails.

xlc:
I don't really get it. Can you share the failing test?

xlc (approved):

ghzlatarev:
> I don't really get it. Can you share the failing test?

Yeah take a look at this test in our repo. It's a little long but essentially register ParaA, ParaB, and relay asset on both ParaA and and ParaB. Also min_xcm_fee for ParaB is registered on ParaA.
Then Alice sends some of ParaB asset to herself on ParaA and mints some of the relay asset to herself on ParaA.
Then Alice sends some of the ParaB tokens back, using relay chain asset as fee.

You can see from the asserts on ParaA and ParaB that things work correctly when `fee_amount` is `50` and `min_xcm_fee` is `10`.

If you change `min_xcm_fee` to 40, the asserts on ParaA (sender chain) will be ok, but the asserts on ParaB (receiver) will fail. This is because the ParaA sovereign account received only `10` of the relay chain asset, when we encoded min_xcm_fee as `fee` when we sent the second `to-reserve` xcm message, so there's not enough to withdraw into Holding and execution stops at the `withdraw_asset` instruction.

zqhxuyuan:
I think your testcase is same as: https://github.com/open-web3-stack/open-runtime-module-library/blob/3576772a896e039abf07c57e1efa305e6c0cddb0/xtokens/src/tests.rs#L717-L778

sender should make sure have enough relaychain token as fee.

ghzlatarev:
Also why do we need to subtract `min-xcm-fee` here https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/src/lib.rs#L558
Take for example case of sending USDT back to Statemine with free execution everywhere to illustrate the point. After both messages get received, my parachain's sovereign account will receive `fee_amount - 2 * min_xcm_fee` and my account will receive `min_xcm_fee` , so 1 `min_xcm_fee` amount seems lost.

zqhxuyuan:
> Also why do we need to subtract `min-xcm-fee` here instead of sending the full fee amout to the relay (and rerouted to parachain) https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/src/lib.rs#L558 Take for example case of sending USDT back to Statemine with free execution everywhere to illustrate the point. After both messages get received, the sender's parachain sovereign account will receive `fee_amount - 2 * min_xcm_fee` and the receiver account will receive `min_xcm_fee` , so 1 `min_xcm_fee` amount seems lost.

If we send all fee_amount to relaychain, then there're none left relaychain token used to send direct to Statemine, But Statemine current only support relaychain token to buy execution fee, so this xcm execution will failed.
Current implementation is split into two xcm, first one is `(fee_amount-min_xcm_fee) KSM` to relaychain then to Statemine, second one is `min_xcm_fee KSM & USDT` direct to Stemine.
If we send `fee_amount KSM` to relaychain then to Statemine(first one), then second one is `USDT` direct to Statemine, it'll failed in Statemine side.

ghzlatarev:
> > Also why do we need to subtract `min-xcm-fee` here instead of sending the full fee amout to the relay (and rerouted to parachain) https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/src/lib.rs#L558 Take for example case of sending USDT back to Statemine with free execution everywhere to illustrate the point. After both messages get received, the sender's parachain sovereign account will receive `fee_amount - 2 * min_xcm_fee` and the receiver account will receive `min_xcm_fee` , so 1 `min_xcm_fee` amount seems lost.
> 
> If we send all fee_amount to relaychain, then there're none left relaychain token used to send direct to Statemine, But Statemine current only support relaychain token to buy execution fee, so this xcm execution will failed. Current implementation is split into two xcm, first one is `(fee_amount-min_xcm_fee) KSM` to relaychain then to Statemine, second one is `min_xcm_fee KSM & USDT` direct to Stemine. If we send `fee_amount KSM` to relaychain then to Statemine(first one), then second one is `USDT` direct to Statemine, it'll failed in Statemine side.

Yes, of-course, thanks for re-explaining the code. For a minute there I though that `assets_to_dest` did not include the `fee_to_dest` and that it was magically being withdrawn into the holding register of Statemine without being included in the `WithdrawAsset` instruction... :D

PR #754, opened 2022-05-29
Introduce `OnNewTokenAccount` and `OnKilledTokenAccount` Callbacks

apopiak:
It would be useful for some features to be able to react to account creation in the tokens pallet that includes the currency for which an account was created.
This PR thus introduces an `OnNewTokenAccount` (and corresponding `OnKilledTokenAccount`) callback to the pallet for other pallets to integrate with.

### DIFF
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -1677,7 +1677,6 @@ impl<T: Config> fungibles::Inspect<T::AccountId> for Pallet<T> {
 		asset_id: Self::AssetId,
 		who: &T::AccountId,
 		amount: Self::Balance,
-		_mint: bool,
 	) -> DepositConsequence {
 		Self::deposit_consequence(who, asset_id, amount, &Self::accounts(who, asset_id))
 	}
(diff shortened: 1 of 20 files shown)


apopiak:
Aside: I was confused I had to include the changes in https://github.com/open-web3-stack/open-runtime-module-library/pull/754/commits/34f50fc5c2399a9820ba69371a809147efb9b354 as they did not seem to have anything to do with this PR. Broken `master`?

xlc:
> Aside: I was confused I had to include the changes in [34f50fc](https://github.com/open-web3-stack/open-runtime-module-library/commit/34f50fc5c2399a9820ba69371a809147efb9b354) as they did not seem to have anything to do with this PR. Broken `master`?

Update your Cargo.toml. That's due to bumping Substrate version

traits/src/currency.rs
```diff
@@ -643,6 +643,24 @@ impl<AccountId, CurrencyId, Balance> OnDust<AccountId, CurrencyId, Balance> for
 	fn on_dust(_: &AccountId, _: CurrencyId, _: Balance) {}
 }
 
+/// Handler for a newly created account
+pub trait OnNewTokenAccount<AccountId, CurrencyId> {
```
xlc, line 647: I don't really like the idea of bunch different traits for each individual purposes.

I think we can just reuse `Happened<(&AccountId, CurrencyId)>` instead.

Otherwise we can have `trait OnAccountUpdated { fn updated }`

apopiak: Have a look. Not sure how I would get around the `who.clone()` induced by the change to `Happened`

tokens/src/lib.rs
```diff
@@ -216,6 +216,12 @@ pub mod module {
 		/// Handler to burn or transfer account's dust
 		type OnDust: OnDust<Self::AccountId, Self::CurrencyId, Self::Balance>;
 
+		/// Handler for when an account was created
+		type OnNewTokenAccount: Happened<(Self::AccountId, Self::CurrencyId)>;
+
+		/// Handler for when an account was created
+		type OnKilledTokenAccount: Happened<(Self::AccountId, Self::CurrencyId)>;
```
zqhxuyuan, line 223: I think we can have a default implementation for those two new trait. maybe use `Nothing`?

xlc:
can you revert the `can_deposit` change and fix the build?

tokens/src/lib.rs
apopiak: There should be a default implementation for `Happened` for `()` which does nothing?

apopiak: Or are you talking about default implementations for traits? I thought that wasn't in stable Rust, yet?

zqhxuyuan: `()` would be ok.

shaunxw (approved):

zjb0807:
Blocked: https://github.com/paritytech/substrate/issues/11608

zjb0807:
@apopiak please merge master to fix the CI.

PR #762, opened 2022-06-06
asset-registry: Pass Asset Id to AuthorityOrigin

NunoAlexandre:
Different AssetId's may require different origin checks, which is currently not supported. 

The changes proposed here make the switch to use `EnsureOriginWithArg` to allow users of this pallet to tailor the origin check to take the `asset_id` itself in consideration when choosing the specific `EnsureOrigin` implementation to be applied.

Fwiw, this feature is a must for our use case at Centrifuge where we need to have a tailored origin check when registering/updating tranche tokens.


asset-registry/src/lib.rs
```diff
@@ -147,7 +145,7 @@ pub mod module {
 			metadata: AssetMetadata<T::Balance, T::CustomMetadata>,
 			asset_id: Option<T::AssetId>,
 		) -> DispatchResult {
-			T::AuthorityOrigin::ensure_origin(origin)?;
+			T::AuthorityOrigin::ensure_origin(origin, &asset_id.clone())?;
```
xlc, line 148: is the clone necessary?

asset-registry/src/mock/para.rs
```diff
@@ -103,11 +104,39 @@ pub struct CustomMetadata {
...
+				.map(|_| ())
+				.map_err(|_| origin),
+
+			// Any other `asset_id` defaults to EnsureRoot
+			_ => EnsureRoot::try_origin(origin),
+		}
+	}
+
+	#[cfg(feature = "runtime-benchmarks")]
+	fn successful_origin(a: &u32) -> OO {
```
shaunxw, line 130:
```suggestion
	fn successful_origin(a: &u32) -> Origin {
```

asset-registry/src/lib.rs
```diff
@@ -88,10 +90,6 @@ pub mod module {
 			asset_id: T::AssetId,
 			metadata: AssetMetadata<T::Balance, T::CustomMetadata>,
 		},
-		SetLocation {
-			asset_id: T::AssetId,
-			location: Box<VersionedMultiLocation>,
-		},
```
shaunxw, line 94: 👍

asset-registry/src/mock/para.rs
NunoAlexandre: my bad, thanks for the fix 👍

NunoAlexandre: my bad, thanks for the fix 👍

NunoAlexandre:
@xlc I will fix the git conflicts and the compilation errors around successful_origin.

xlc (approved):

PR #767, opened 2022-06-12
asset-registry: And and Implement `Inspect` + `Mutate` traits

NunoAlexandre:
This allows for loosely coupling of the `orml_asset_registry` pallet instead of forcing other pallets into a tight dependency.

This is particularly helpful for us at Centrifuge where we need to pass the asset-registry to a few pallets to query asset metadata and register tranche tokens.

### DIFF
--- a/asset-registry/src/impls.rs
+++ b/asset-registry/src/impls.rs
@@ -1,7 +1,9 @@
-use crate::{module::*, AssetMetadata};
+use crate::module::*;
 use frame_support::{log, pallet_prelude::*, weights::constants::WEIGHT_PER_SECOND};
 use orml_traits::{
-	asset_registry::{AssetProcessor, FixedConversionRateProvider, WeightToFeeConverter},
+	asset_registry::{
+		AssetMetadata, AssetProcessor, FixedConversionRateProvider, Inspect, Mutate, WeightToFeeConverter,
+	},
 	GetByKey,
 };
 use sp_runtime::FixedPointNumber;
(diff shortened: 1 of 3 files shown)


xlc:
Looks fine. Testing failing.

NunoAlexandre:
@xlc It's ready for another round

asset-registry/src/impls.rs
```diff
@@ -169,3 +171,54 @@ impl<T: Config> GetByKey<T::AssetId, T::Balance> for ExistentialDeposits<T> {
...
+		Pallet::<T>::do_register_asset(metadata, asset_id)
+	}
+
+	fn update_asset(
+		asset_id: Self::AssetId,
+		decimals: Option<u32>,
+		name: Option<Vec<u8>>,
+		symbol: Option<Vec<u8>>,
+		existential_deposit: Option<Self::Balance>,
+		location: Option<Option<xcm::VersionedMultiLocation>>,
```
zqhxuyuan, line 211: use `VersionedMultiLocation` here, and import it at header.

zqhxuyuan: also `use xcm::v2::prelude::*;` can be change to `xcm::latest::prelude::*`

NunoAlexandre:
@zqhxuyuan I just pushed one extra commit to `pub use` the `AssetMetadata` struct from the pallet itself to keep things the way they were for users.

zqhxuyuan (approved):

PR #771, opened 2022-06-13
rewards split

dzmitry-lahoda:
so we need to split users position. he can do that by transferring share with rewards

### DIFF
--- a/rewards/src/lib.rs
+++ b/rewards/src/lib.rs
@@ -10,7 +10,6 @@ use orml_traits::RewardHandler;
 use scale_info::TypeInfo;
 use sp_core::U256;
 use sp_runtime::{
-	helpers_128bit::multiply_by_rational,
 	traits::{AtLeast32BitUnsigned, MaybeSerializeDeserialize, Member, Saturating, UniqueSaturatedInto, Zero},
 	FixedPointOperand, RuntimeDebug, SaturatedConversion,
 };
@@ -327,7 +326,7 @@ impl<T: Config> Pallet<T> {
 	///
 	/// Similar too claim and add 2 shares later, but does not requires pool
 	/// inflation and is more efficient.
-	pub fn transfer_share(
+	pub fn transfer_share_and_rewards(
 		who: &T::AccountId,
 		pool: &T::PoolId,
 		move_share: T::Share,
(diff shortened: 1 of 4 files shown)


rewards/src/lib.rs
```diff
@@ -10,6 +10,7 @@ use orml_traits::RewardHandler;
 use scale_info::TypeInfo;
 use sp_core::U256;
 use sp_runtime::{
+	helpers_128bit::multiply_by_rational,
```
zqhxuyuan, line 13: use `multiply_by_rational_with_rounding` instead of `multiply_by_rational`.

dzmitry-lahoda: hm, seems it needs to upgrade polkadot deps. is not it? should than this PR merged after update OR can live with multiply_by_rational OR can make it work with u256?

dzmitry-lahoda: not sure, used u256 directly

rewards/src/lib.rs
```diff
@@ -315,6 +319,53 @@ impl<T: Config> Pallet<T> {
...
+	) -> DispatchResult {
+		SharesAndWithdrawnRewards::<T>::mutate(pool, other, |increased_share| {
+			let (increased_share, increased_rewards) = increased_share;
+			SharesAndWithdrawnRewards::<T>::mutate_exists(pool, who, |share| {
+				if let Some((share, rewards)) = share {
+					if move_share < *share {
+						for (reward_currency, balance) in rewards {
+							let move_balance = U256::from(balance.to_owned().saturated_into::<u128>())
+								* U256::from(move_share.to_owned().saturated_into::<u128>())
+								/ U256::from(share.to_owned().saturated_into::<u128>());
```
xlc, line 343: need some comment to explain why this is safe

dzmitry-lahoda: explained some, so i feel back about that Share is not capped to some number so and saturated used in many places (thins will get broken if share is > u128, not only in this method so). i feel like Share needs more traits. they are not related mbr or u256, but converte into u128 said first.

dzmitry-lahoda: so this method seems as safe as others, until replace all sat_into_u128 with try_into or extend share traits. will close comment for now if that is ok

rewards/src/lib.rs
```diff
@@ -41,8 +41,17 @@ where
 
 pub use module::*;
 
+#[macro_export]
+macro_rules! ensure_some {
```
xlc, line 45: this is unnecessary

just `let x = x.ok_or(err)?;`

dzmitry-lahoda:
cc @xlc @zqhxuyuan

xlc (approved):

PR #785, opened 2022-06-24
NonReserve case support teleport instead of reserve transfer

zqhxuyuan:
Current [`fee_reserve != non_fee_reserve`](https://github.com/open-web3-stack/open-runtime-module-library/blob/27e3272d322dcdd915f0fc2002032e8d53a46523/xtokens/src/lib.rs#L537) only used in parachain transfer back asset back to Statemine, The first xcm we can change to teleport.

### DIFF
--- a/xcm-support/src/currency_adapter.rs
+++ b/xcm-support/src/currency_adapter.rs
@@ -145,6 +145,12 @@ impl<
 		DepositFailureHandler,
 	>
 {
+	fn check_in(_origin: &MultiLocation, _what: &MultiAsset) {}
+
+	fn can_check_in(_origin: &MultiLocation, _what: &MultiAsset) -> Result {
+		Ok(())
+	}
+
 	fn deposit_asset(asset: &MultiAsset, location: &MultiLocation) -> Result {
 		match (
 			AccountIdConvert::convert_ref(location),
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -571,6 +571,7 @@ pub mod module {
 					&dest,
 					Some(T::SelfLocation::get()),
 					dest_weight,
+					true,
 				)?;
 
 				// Second xcm send to dest chain.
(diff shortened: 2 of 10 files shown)


xcm-support/src/currency_adapter.rs
```diff
@@ -145,6 +145,12 @@ impl<
 		DepositFailureHandler,
 	>
 {
+	fn check_in(_origin: &MultiLocation, _what: &MultiAsset) {}
```
zqhxuyuan, line 148: Need to align with polkadot CurrencyAdapter: https://github.com/paritytech/polkadot/blob/b0722401e1d66e6b5470995bf748467ea3fb92dd/xcm/xcm-builder/src/currency_adapter.rs#L97-L134

xlc: why do we need to implement those? I don't think we will ever support teleport operation locally

zqhxuyuan: Oh yes, only statemine need this `CurrencyAdapter` implementation, but in our testcase both sender and receiver are using same `MultiCurrencyAdapter`. I'll try let sender use `MultiCurrencyAdapter` and receiver use `CurrencyAdapter`.

zqhxuyuan: If `ParaC` using polkadot CurrencyAdapter, there'll be execution error in testcase, So I add another `MultiTeleportCurrencyAdapter` inside mock, this `MultiTeleportCurrencyAdapter` add default `check_in` and `can_check_in` implemenats. It's used only in testcase.

xtokens/src/lib.rs
```diff
@@ -571,6 +574,7 @@ pub mod module {
 					&dest,
 					Some(T::SelfLocation::get()),
 					dest_weight,
+					true,
```
zqhxuyuan, line 577: See the comment above, I just set `true` here, then this first xcm will always using `teleport`. Do we need use configuration setting like `TrustedTeleporters` to decide whether use teleport or not?

PR #787, opened 2022-07-01
Reanchor recipient in `to-non-reserve transfer`

ghzlatarev:
* In the case of sending back assets to statemine the recipient is overriden with `SelfLocation` and when this self location is in relative view, it will not be interpreted correctly on the receiver chain.
* So to support relative views in this case, we can simply reanchor the recipient from the point of view of the destination chain. However only in the case of the recipient being (0, Here)
* Added a test case for this - `sending_sibling_asset_to_reserve_sibling_with_relay_fee_works_with_relative_self_location`

Signed-off-by: Georgi Zlatarev <georgi.zlatarev@manta.network>

### DIFF
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -706,6 +706,12 @@ pub mod module {
 				}
 			}
 
+			let mut reanchored_recipient = recipient.clone();
+			if recipient == MultiLocation::here() {
+				let ancestry = T::LocationInverter::ancestry();
+				reanchored_recipient.reanchor(&dest, &ancestry);
+			}
+
 			if !use_teleport {
 				Ok(Xcm(vec![
 					WithdrawAsset(assets.clone()),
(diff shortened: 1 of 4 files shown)


xtokens/src/lib.rs
```diff
@@ -706,6 +706,14 @@ pub mod module {
 				}
 			}
 
+			let mut reanchored_recipient = recipient.clone();
+			if recipient == MultiLocation::here() {
+				let ancestry = T::LocationInverter::ancestry();
+				let _ = reanchored_recipient
+					.reanchor(&dest, &ancestry)
+					.map_err(|_| DispatchError::Other("Cannot reanchor recipient"));
```
zqhxuyuan, line 714: the error type should be `Error::<T>::CannotReanchor`.

zqhxuyuan: Also can you put this logic into `do_transfer_multiassets` before invoke `execute_and_send_reserve_kind_xcm`. I think this is a special process only for transfer asset back to statemine/t.

zqhxuyuan: Also modified last two paragraph in https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/README.md

ghzlatarev: Thanks for the review, please take a look at the resolutions.

zqhxuyuan (approved):

PR #793, opened 2022-07-16
update polkadot-v0.9.26

zqhxuyuan:


traits/src/location.rs
```diff
@@ -82,7 +82,13 @@ pub trait RelativeLocations {
 
 impl RelativeLocations for MultiLocation {
 	fn sibling_parachain_general_key(para_id: u32, general_key: Vec<u8>) -> MultiLocation {
-		MultiLocation::new(1, X2(Parachain(para_id), GeneralKey(general_key)))
+		MultiLocation::new(
+			1,
+			X2(
+				Parachain(para_id),
+				GeneralKey(general_key.to_vec().try_into().expect("less than length limit; qed")),
```
xlc, line 89: I don't think this is safe. We should change the `general_key` type

xlc (approved):

Kailai-Wang:
Sorry, but is the deletion of `polkadot-v0.9.26` branch intended? Shall we depend on the master branch in case we want to upgrade the deps to v0.9.26?

xlc:
sorry @Kailai-Wang restored.

Kailai-Wang:
All good, thanks @xlc !

PR #799, opened 2022-08-30
update polkadot-v0.9.28

1xstj:
Bump dependencies to polkadot-v0.9.28


authority/src/tests.rs
```diff
@@ -80,13 +80,13 @@ fn schedule_dispatch_at_work() {
 				delay: 1,
 				origin: Box::new(OriginCaller::system(RawOrigin::Root)),
 			}),
-			index: 1,
+			index: 0,
```
shaunxw, line 83: Could you please give some details why the index is changed?

1xstj: Similarly for the other test aswell.

zjb0807: https://github.com/open-web3-stack/open-runtime-module-library/blob/27115ac7745662c2134862fdd8fc77edf6c9ad04/authority/src/lib.rs#L259-L273

There is no `[transactional]` here, the failed tests are not rolled back

zjb0807 (approved):

PR #800, opened 2022-09-02
Remove transactional

### LINKED ISSUE #784: Remove `transactional` on extrinsic
should fix all `assert_noop` in tests:

https://github.com/paritytech/substrate/pull/11546#issuecomment-1141813767

zjb0807:
Closes: #784

### DIFF
--- a/payments/src/lib.rs
+++ b/payments/src/lib.rs
@@ -72,8 +72,8 @@ pub mod pallet {
 		weights::WeightInfo,
 	};
 	use frame_support::{
-		dispatch::DispatchResultWithPostInfo, fail, pallet_prelude::*, storage::bounded_btree_map::BoundedBTreeMap,
-		traits::tokens::BalanceStatus,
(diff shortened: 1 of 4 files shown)


payments/src/lib.rs
```diff
@@ -544,7 +536,6 @@ pub mod pallet {
 		/// The function will create a new payment. The fee and incentive
 		/// amounts will be calculated and the `PaymentDetail` will be added to
 		/// storage.
-		#[require_transactional]
```
xlc, line 547: we can still keep `require_transactional`

tokens/src/lib.rs
```diff
@@ -2135,7 +2135,6 @@ where
 }
 
 impl<T: Config> TransferAll<T::AccountId> for Pallet<T> {
-	#[transactional]
```
shaunxw, line 2138: I think we can keep `transactional` for trait implementations, which could be used in `on_initialize/finalize` which are not `transactional` on default.

xlc (approved):

shaunxw (approved):

PR #804, opened 2022-09-20
make clear_prefix count optional

ermalkaleci:


bencher/test/src/lib.rs
```diff
@@ -77,12 +77,12 @@ pub mod pallet {
 
 		#[orml_weight_meter::weight(0)]
 		pub(crate) fn remove_all_bar() {
-			Bar::<T>::remove_all(None);
+			_ = Bar::<T>::clear(10, None);
```
zjb0807, line 80: should it be `_ = Bar::<T>::clear(u32::max_value(), None)`?

ermalkaleci: no, it's just test

ermalkaleci: I made it optional because now it's required to pass a number so it's optional if you want to count that as writes

PR #815, opened 2022-10-14
feat(tokens): OnDeposit, OnTransfer, OnSlash hooks

savudani8:
Adds configurable hooks to the Tokens pallet, for functions which emit `Event::Deposited`, `Event::Transfer`, `Event::Slashed`. A hook call is not added to `slash_reserved_named(...)`, because that function calls `slash_reserved(...)` so the hook is run there.

These hooks enable tokenizing "positions" which require lazy operations to be executed each time they are modified. An example would be tokenizing one's nomination / stake. For Alice to send her nominated position (the right to claim the nominated tokens) to Bob, her nomination rewards have to first be collected. In this example, the transfer hook would collect the sender's rewards before the actual transfer occurs.

### DIFF
--- a/tokens/src/lib.rs
+++ b/tokens/src/lib.rs
@@ -66,7 +66,7 @@ use sp_std::{cmp, convert::Infallible, marker, prelude::*, vec::Vec};
 
 use orml_traits::{
 	arithmetic::{self, Signed},
-	currency::TransferAll,
+	currency::{OnDeposit, OnSlash, OnTransfer, TransferAll},
 	BalanceStatus, GetByKey, Happened, LockIdentifier, MultiCurrency, MultiCurrencyExtended, MultiLockableCurrency,
 	MultiReservableCurrency, NamedMultiReservableCurrency, OnDust,
 };
@@ -173,6 +173,8 @@ pub use module::*;
 
 #[frame_support::pallet]
 pub mod module {
+	use orml_traits::currency::{OnDeposit, OnSlash, OnTransfer};
+
 	use super::*;
 
 	#[pallet::config]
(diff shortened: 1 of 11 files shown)


tokens/src/lib.rs
```diff
@@ -890,6 +901,7 @@ impl<T: Config> Pallet<T> {
 		amount: T::Balance,
 		existence_requirement: ExistenceRequirement,
 	) -> DispatchResult {
+		T::OnTransfer::on_transfer(currency_id, from, to, amount)?;
```
xlc, line 904: move this after the early return. on the case no event is emitted, no hook should be called.

traits/src/currency.rs
```diff
@@ -657,3 +657,34 @@ impl<AccountId> TransferAll<AccountId> for Tuple {
...
+}
+
+impl<AccountId, CurrencyId, Balance> OnDeposit<AccountId, CurrencyId, Balance> for () {
+	fn on_deposit(_: CurrencyId, _: &AccountId, _: Balance) -> DispatchResult {
+		Ok(())
+	}
+}
+
+/// Hook to run before transferring from an account to another.
+pub trait OnTransfer<AccountId, CurrencyId, Balance> {
```
xlc, line 682: Do you intent to allow OnTransfer to be able to cancel the transfer?

xlc (changes requested):
need more tests

traits/src/currency.rs
savudani8: Ideally I wanted all of these hooks to be able to fail, in consequence cancelling the transfer / deposit / slash. The trait bound on `slash()` prevents such an implementation unfortunately, but the other two (on_deposit and on_transfer) are allowed to return a `Result` so that's what they do.

xlc (approved):

zjb0807 (approved):
LGTM

PR #819, opened 2022-10-22
#xcm if assets is of lower decimal then we want then pow it local

dzmitry-lahoda:
# motivation

- we want to normalize all arriving xcm assets to specific decimals

# implementation 

- default impl is no normalization
- only safe path of rising precision is used (xcm asset use reserver precision, local use increased)

### DIFF
--- a/asset-registry/src/mock/para.rs
+++ b/asset-registry/src/mock/para.rs
@@ -17,7 +17,7 @@ use orml_traits::{
 	location::{AbsoluteReserveProvider, RelativeReserveProvider},
 	parameter_type_with_key, FixedConversionRateProvider, MultiCurrency,
 };
-use orml_xcm_support::{IsNativeConcrete, MultiCurrencyAdapter, MultiNativeAsset};
+use orml_xcm_support::{EqualDecimalsNormalizer, IsNativeConcrete, MultiCurrencyAdapter, MultiNativeAsset};
 use pallet_xcm::XcmPassthrough;
 use polkadot_parachain::primitives::Sibling;
 use sp_core::H256;
@@ -184,6 +184,7 @@ pub type LocalAssetTransactor = MultiCurrencyAdapter<
 	CurrencyId,
 	CurrencyIdConvert,
 	(),
+	EqualDecimalsNormalizer,
 >;
 
 pub type XcmRouter = ParachainXcmRouter<ParachainInfo>;
--- a/xcm-support/Cargo.toml
+++ b/xcm-support/Cargo.toml
@@ -10,6 +10,8 @@ edition = "2021"
 [dependencies]
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false }
 
+num-traits = { version = "0.2.14", default-features = false }
+
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.29" }
 sp-runtime = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.29" }
 frame-support = { git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.29" }
--- a/xcm-support/src/currency_adapter.rs
+++ b/xcm-support/src/currency_adapter.rs
@@ -4,6 +4,9 @@ use sp_runtime::{
 	traits::{Convert, MaybeSerializeDeserialize, SaturatedConversion},
 	DispatchError,
 };
+
+use num_traits::pow::checked_pow;
+
 use sp_std::{
 	cmp::{Eq, PartialEq},
 	fmt::Debug,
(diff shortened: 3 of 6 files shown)


xlc (changes requested):
This isn't the right place to handle decimal conversion, which usually is a bad idea anyway.
If you really want to do it, you should just implement a currency wrapper that implements `MultiCurrency` trait but does the decimal conversion and delegate the converted values to the underlying implementation.

`Mapper` could also be used depends on your use case
https://github.com/open-web3-stack/open-runtime-module-library/blob/2c48b626addafbea0227c068c9ab0fc16666b1b5/tokens/src/impls.rs#L126

dzmitry-lahoda:
> which usually is a bad idea anyway.

i feel it bad too(we just decided to do so). But why? I would be glad to have arguments against coversion so I do not have to do it:)

> MultiCurrency
Yeah. I see. There is revenue from holding (assettrap) which also must be handled. So I have to handle DepositFailureHandler  and Match. It seems brittle, so yeah - need to do MultiCurrency mapper.

xlc:
one of the reason is that it will be impossible to support unknown tokens because you don’t know the decimals

dzmitry-lahoda:
Likely we will not do conversion, so I still believe that gating decimals should be as close as possible to bridge - so xcm-support is good place. Because then you either can handle parts xtokens for UI decimals and handle imbapances for XCM bridge. handling bridge imbalance does not feels like role of assets mappers.

```
Dzmitry Lahoda
  [4 hours ago](https://composablefinance.slack.com/archives/C03LBRRMU4U/p1666539280843609?thread_ts=1666093938.150289&cid=C03LBRRMU4U)
In XCM area, there is parity xcm-executor, which is VM interpreter (and there are pallet-xcm and orml-xtokens to ensure secured extrinsics to this executor). These do not assume any decimals and have no information about precision.
These call or call back into assets system to transfer tokens, drop unknown tokens, withdraw tokens, burn tokens, etc. That integration based on orml xcm-support. It is not aware about decimals too.
And finally, xcm-support calls our xcm <-> assets integration layer which have access to foreign assets decimals.
Solution suggested by acala devs, wrap MultiCurrency or/and other relevant traits(or other solution fork xcm-support to hook directly into xcm-executor) which callback from xcm-support. In that wrapper any call would assume decimals are original decimals of origin of asset. And will multiply always to increase precision or divide to reduce accuracy.
During division we have to handle cases when you want to withdraw and send to other chain, you have to round up reduction (because our reserve must me the same or less than tracked on origin chain).  So if I send 12345. You send XCM to other chain with 12345, but locally you withdraw 124. When XCM receives and wants to put 12345 on our chain. You have to decrease the minted amount to 123.  This kind of safe solution but will accrue disbalance over time. So will need to track imbalances somehow (and who owns imbalances?).
And there is FE issue to tackel. It has to send origin decimals amount. So XCM api in our FE will be working with original decimals, while other on chain parts will work with our chain decimals.
That can be handled by more deep modification/extensions/forks of xcm-executor/pallet-xcm/orml-xtokens.
So avoding having decimals mappings removes complexity from implementation of this stuff. If higher precision of some token would be needed ever, can have pallet which just does that and make it part of dex router and handle inmbalances local.
I know that Oracle reports prices. Oracle should report prices in some decimals. Because there are not transfers, no imbalances and readonly regarding mint and burn. It is easy to raise or down precisions using neares number rounding.
Other place where we do assume decimals is tests (example Angular tests). So I we tried to encode that not like 10^6, but USDT::UNIT or similar patterns.
Tests in XCM simulator do not assume any specific defaul decimals and work with relation of decimals and rations all the time.
composable
+1::skin-tone-6
raised_hands::skin-tone-6





Dzmitry Lahoda
  [4 hours ago](https://composablefinance.slack.com/archives/C03LBRRMU4U/p1666539326859859?thread_ts=1666093938.150289&cid=C03LBRRMU4U)
I have not seen assumption of decimals in Staking. I do not know status of Pablo.


Dzmitry Lahoda
  [1 hour ago](https://composablefinance.slack.com/archives/C03LBRRMU4U/p1666548765930459?thread_ts=1666093938.150289&cid=C03LBRRMU4U)
One case when one may want automatic decimal increase if origin unit is 6 decimals. Price of unit is 1M USD. In this case minimal free could be taken is 1 which is a lot. So that seems unlikely as currencies are 6 decimals, 9 or more. And there is no 1M currency (when bitcoin will reach such, it it is >= 9 decimals)


Dzmitry Lahoda
  [1 hour ago](https://composablefinance.slack.com/archives/C03LBRRMU4U/p1666548775406089?thread_ts=1666093938.150289&cid=C03LBRRMU4U)
``

xlc:
I am not aware any existing bridges cares about decimals. Decimals is really a business logic thing. For many use cases, we don't need decimals at all. Therefore I think it is a bad idea to bake it into the low level implementation.

dzmitry-lahoda:
Statemine assets allow to create assets without decimals.

Issue #824, opened 2022-10-27
consider using checked_add in orml_tokens's do_deposit


enthusiastmartin:
orml-tokens 

do_deposit performs the following:

```rust
    let new_total_issuance = Self::total_issuance(currency_id)
	.checked_add(&amount)
	.ok_or(ArithmeticError::Overflow)?;
    if change_total_issuance {
	TotalIssuance::<T>::mutate(currency_id, |v| *v = new_total_issuance);
    }
    account.free += amount;
```

I'd suggest to use safe math for updating `account.free` like so:
```
   account.free = account.free.checked_add(&amount).ok_or(ArithmeticError::Overflow)?;
```

I am aware that safety is or should be guaranteed by previous check of adding amount to total issuance as free should be <= total_issuance. 

However, our audit brought this up while auditing our pallet and would recommend to change this to prevent surprising overflows when for some reason total issuance is updated incorrectly, albeit very unlikely.

PR #828, opened 2022-10-31
Add checked_* and saturating_*

zjb0807:
Closes: #824


xlc (commented):
Should use defensive traits https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/misc.rs

tokens/src/lib.rs
```diff
@@ -603,15 +603,17 @@ pub mod module {
 
 				if new_total > old_total {
 					TotalIssuance::<T>::try_mutate(currency_id, |t| -> DispatchResult {
+						// this sub can't underflow but just to be defensive here.
```
xlc, line 606: no need to have those comments. the defensive methods means exactly this

xlc (approved):

xlc (approved):

PR #832, opened 2022-11-07
Applied fixes for unbalanced traits

PoisonPhang:
Applied the fixes found in #823 to branch `polkadot-v0.9.27`

Our runtime still depend on Polkadot v0.9.27. However, this fix is necessary our runtime to function as expected.

PR #833, opened 2022-11-07
Backport fixes to v0.9.27

xlc:
replaces #832

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,11 +4,13 @@ on:
   pull_request:
     branches:
       - master
+      - polkadot-**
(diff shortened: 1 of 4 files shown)


PoisonPhang:
Was there a problem with #832 that could not be addressed within the existing PR?

Could we consider splitting up the many test cases into their own unit tests? The combined test for the entire trait is not friendly for debugging, less maintainable, and doesn't do a good job at explaining what it is testing.

xlc:
A backport should be a backport. If you want to improve the tests, please do it on master and backport it to release branches. I don't want to see new developments on release branches.

PR #834, opened 2022-11-08
Token hook refactoring; Posthooks for deposit and transfer

savudani8:
First commit: Refactors all hooks in the `tokens` pallet (also called handlers in the code) into a single associated type that provides a default implementation. This clears the way for adding new hooks in a backwards-compatible way.

Second commit: Adds post-hooks to the deposit and transfer functions, such that consumers can set custom actions to be run immediately after the functions complete. The previous `OnDeposit` and `OnTransfer` hooks have been renamed to `DepositPreHook` and `TransferPreHook` respectively, because they run before the actual action.


tokens/src/lib.rs
```diff
@@ -765,11 +751,11 @@ impl<T: Config> Pallet<T> {
 				// Ignore the result, because if it failed then there are remaining consumers,
 				// and the account storage in frame_system shouldn't be reaped.
 				let _ = frame_system::Pallet::<T>::dec_providers(who);
-				T::OnKilledTokenAccount::happened(&(who.clone(), currency_id));
+				<T::CurrencyHooks as CurrencyMutationHooks<T::AccountId, T::CurrencyId, T::Balance>>::OnKilledTokenAccount::happened(&(who.clone(), currency_id));
```
shaunxw, line 754: you can define a `MutationHooksOf<T>` type alias to avoid repeating

traits/src/currency.rs
```diff
@@ -690,3 +690,40 @@ impl<AccountId, CurrencyId, Balance> OnTransfer<AccountId, CurrencyId, Balance>
 		Ok(())
 	}
 }
+
+pub trait CurrencyMutationHooks<AccountId, CurrencyId, Balance> {
```
shaunxw, line 694:
```suggestion
pub trait MutationHooks<AccountId, CurrencyId, Balance> {
```

traits/src/currency.rs
```diff
@@ -690,3 +690,40 @@ impl<AccountId, CurrencyId, Balance> OnTransfer<AccountId, CurrencyId, Balance>
...
+	/// Handler to burn or transfer account's dust
+	type OnDust: OnDust<AccountId, CurrencyId, Balance>;
+
+	/// Hook to run before slashing an account.
+	type OnSlash: OnSlash<AccountId, CurrencyId, Balance>;
+
+	/// Hook to run before depositing into an account.
+	type DepositPreHook: OnDeposit<AccountId, CurrencyId, Balance>;
```
shaunxw, line 702: would be good to keep naming consistent, like `type PreDeposit: ...`

tokens/src/lib.rs
savudani8: I'm not sure how that would work exactly? The error I'm getting with the type alias is
```rust
error[E0223]: ambiguous associated type
   --> tokens/src/lib.rs:941:3
    |
941 |         MutationHooksOf::<T>::TransferPostHook::on_transfer(
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use fully-qualified syntax: `<<T as module::Config>::CurrencyHooks as Trait>::TransferPostHook` 
```

Which happens because of [this issue](https://github.com/rust-lang/rust/issues/38078).

shaunxw (approved):

savudani8:
@shaunxw do you have any tips for fixing the coverage error? I'm not sure if my change is the one causing it

xlc:
We need to update the auth token. It is unrelated to this PR.

xlc:
Try merge master

xlc (approved):

PR #839, opened 2022-11-14
fix(tokens): PostDeposit hook

savudani8:
The `PostDeposit` hook was running inside `try_mutate_account`, so the user's balance was still equal to the pre-deposit amount. This caused certain post-deposit actions to fail.

#### Reason why this was not caught in the tests
The post-hook tests just increment a variable, so there are no real side effects. I thought about adding balance-changing posthooks inside the mock runtime, but those hooks would affect the results of many other tests. I don't think there's an easy way to use multiple mock runtimes in unit tests either.

### DIFF
--- a/tokens/src/mock.rs
+++ b/tokens/src/mock.rs
@@ -331,9 +331,9 @@ impl<T: Config> OnDeposit<T::AccountId, T::CurrencyId, T::Balance> for PostDepos
 		ON_DEPOSIT_POSTHOOK_CALLS.with(|cell| *cell.borrow_mut() += 1);
 		let account_balance: AccountData<T::Balance> =
 			tokens::Pallet::<T>::accounts::<T::AccountId, T::CurrencyId>(account_id.clone(), currency_id);
-		ensure!(
+		assert!(
 			account_balance.free.ge(&amount),
-			DispatchError::Other("Posthook must run after the account balance is updated.")
+			"Posthook must run after the account balance is updated."
 		);
 		Ok(())
 	}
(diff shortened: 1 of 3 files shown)


savudani8:
> one why to write tests for this is to instead of just record call count, record the hook parameters in a vector. then you can assert the vector for expected number of calls & expected called values

I tried implementing your suggestion but realised that call parameters were correct, it was the timing of the call that was wrong.

As such, I added checks in the mock posthooks which enforce that the balance of the account is at least `amount`. I was then able to test the edge case that this PR fixes.

xlc:
In that case you can read the account data and have it saved to the vector for assertion as well. But I am ok with your current approach.

xlc (approved):

tokens/src/mock.rs
```diff
@@ -327,8 +327,14 @@ impl<T: Config> PreDeposit<T> {
 
 pub struct PostDeposit<T>(marker::PhantomData<T>);
 impl<T: Config> OnDeposit<T::AccountId, T::CurrencyId, T::Balance> for PostDeposit<T> {
-	fn on_deposit(_currency_id: T::CurrencyId, _account_id: &T::AccountId, _amount: T::Balance) -> DispatchResult {
+	fn on_deposit(currency_id: T::CurrencyId, account_id: &T::AccountId, amount: T::Balance) -> DispatchResult {
 		ON_DEPOSIT_POSTHOOK_CALLS.with(|cell| *cell.borrow_mut() += 1);
+		let account_balance: AccountData<T::Balance> =
+			tokens::Pallet::<T>::accounts::<T::AccountId, T::CurrencyId>(account_id.clone(), currency_id);
+		ensure!(
```
xlc, line 334: actually, this should be an assert instead of ensure

xlc (changes requested):
should use assert in the hooks instead of ensure

xlc (approved):

PR #841, opened 2022-11-15
refactor: make weight argument in xtokens transfers optional

### LINKED ISSUE #838: Replace weight argument in xtokens::transfer with max_fee
Related to https://github.com/open-web3-stack/open-runtime-module-library/issues/656.

In practice, I believe people just put some arbitrary high amount for the weight in `xtokens::transfer`. To me it does make sense to have this as an argument at all. Users do not care how much weight is used on the destination chain - it is simply a representation of computation time on the remote chain. As such, it makes more sense to me to either have the `weight` argument be optional, or to remove it altogether. If not supplied, the code would set the limit to `Unlimited`.

If anything, users may care about how many fees they pay. So we could consider adding an optional `max_fee` argument to `xtokens::transfer`. Note that the `BuyExecution` instruction takes both a max_fee and a max_weight argument. However, even the use `max_fee` is dubious. If the fee is insufficient, presumably the transferred tokens get trapped and the user can then recover them using a `ClaimAsset` instruction. However, the user would still end up paying the fees to pay for execution of this new message..

Interested in hearing your thoughts about this.

sander2:
Closes #838.

Make the weight argument in xtokens transfers optional since in practice it's very rare to set a specific limit - most of the time people just set some arbitrary high value. With this change, you can pass `None` to set the BuyExecution limit to `Unlimited`, which is cleaner and less fragile.

### DIFF
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -205,7 +205,7 @@ pub mod module {
 			currency_id: T::CurrencyId,
 			amount: T::Balance,
 			dest: Box<VersionedMultiLocation>,
-			dest_weight: Weight,
+			dest_weight: WeightLimit,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 			let dest: MultiLocation = (*dest).try_into().map_err(|()| Error::<T>::BadVersion)?;
@@ -229,7 +229,7 @@ pub mod module {
 			origin: OriginFor<T>,
 			asset: Box<VersionedMultiAsset>,
 			dest: Box<VersionedMultiLocation>,
-			dest_weight: Weight,
+			dest_weight: WeightLimit,
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 			let asset: MultiAsset = (*asset).try_into().map_err(|()| Error::<T>::BadVersion)?;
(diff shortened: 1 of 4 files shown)


shaunxw (approved):
LGTM

xtokens/src/lib.rs
```diff
@@ -763,16 +763,19 @@ pub mod module {
...
 				.map_err(|_| Error::<T>::CannotReanchor)?;
-			Ok(BuyExecution {
-				fees,
-				weight_limit: WeightLimit::Limited(weight),
-			})
+			let weight_limit = if let Some(limit) = weight {
+				WeightLimit::Limited(limit)
+			} else {
+				WeightLimit::Unlimited
+			};
```
NunoAlexandre, line 776: `nit` but we could make this a little more succint and elegant imo:

```suggestion
			let weight_limit = limit
			  .map(|x| WeightLimit::Limited(x))
			  .unwrap_or(WeightLimit::Unlimited);
```

NunoAlexandre (commented):
This looks good but we could also choose to make `dest_weight` of type `WeightLimit` instead of `Option<Weight>` and then having to convert it to `WeightLimit`; that also makes the API more readable by passing `WeightLimit::Unlimited` instead of `None`.

Very happy to have this in either way 🚀

sander2:
> This looks good but we could also choose to make `dest_weight` of type `WeightLimit` instead of `Option<Weight>` and thus having to convert it to `WeightLimit`; doing so would also make the API more readable by passing `WeightLimit::Unlimited` instead of `None`.

Yea that makes sense, I'll try to update the pr later today

NunoAlexandre (approved):
💯

xlc (approved):

PR #845, opened 2022-11-21
prevent nested DelayedOrigin

xlc:
Fixes https://substrate.stackexchange.com/questions/5970/implement-maxencodedlen-for-nested-origin

One issue is that this make decode invalid but it is still possible to construct nested DelayedOrigin normally.
This means the storage will be inconsistent if someone is ever successfully wrote origin containing nested DelayedOrigin. The read will fail.

Maybe the safer way is not making decoding nested DelayedOrigin invalid. Just guarded enough so ensure it cannot be constructed but even it does, it is not a state consistency error but _just_ a PoV size estimation issue.

### DIFF
--- a/authority/src/tests.rs
+++ b/authority/src/tests.rs
@@ -698,17 +698,3 @@ fn origin_max_encoded_len_works() {
 	assert_eq!(DelayedOrigin::<u32, OriginCaller>::max_encoded_len(), 22);
 	assert_eq!(OriginCaller::max_encoded_len(), 27);
 }
-
-#[test]
-fn nested_delayed_origin_is_invalid() {
-	let orgin = DelayedOrigin::<u64, OriginCaller> {
-		delay: 1u64,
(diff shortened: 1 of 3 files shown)


authority/src/tests.rs
```diff
@@ -691,3 +692,23 @@ fn trigger_old_call_should_be_free_and_operational() {
...
+	let orgin = DelayedOrigin::<u64, OriginCaller> {
+		delay: 1u64,
+		origin: Box::new(OriginCaller::Authority(DelayedOrigin::<u64, OriginCaller> {
+			delay: 1u64,
+			origin: Box::new(OriginCaller::system(frame_system::RawOrigin::Root)),
+		})),
+	};
+
+	let encoded = orgin.encode();
+	assert!(OriginCaller::decode(&mut &encoded[..]).is_err());
```
zjb0807, line 713:
```suggestion
	assert!(DelayedOrigin::<u64, OriginCaller>::decode(&mut &encoded[..]).is_err());
```

shaunxw (approved):

PR #854, opened 2022-11-27
impl rate limit for xtokens

wangjj9219:


rate-limit/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+		Unlimited,
+		/// is_allowed check return false always.
+		NotAllowed,
+	}
+
+	/// Match rules to fitler key is in bypass whitelist.
+	#[derive(PartialOrd, Ord, PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+	pub enum KeyFilter {
+		/// If the encoded key is equal to the vec, the key is in whitelist.
+		Match(Vec<u8>),
```
xlc, line 65: need to use BoundedVec

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+
+				*remainer_quota
+			})
+		}
+	}
+
+	impl<T: Config> RateLimiter for Pallet<T> {
+		type RateLimiterId = T::RateLimiterId;
+
+		fn bypass_limit(limiter_id: Self::RateLimiterId, key: impl Encode) -> bool {
```
xlc, line 367:
```suggestion
		fn can_bypass_limit(limiter_id: Self::RateLimiterId, key: impl Encode) -> bool {
```

or `is_whitelisted`

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+								blocks_count,
+								quota_increment,
+								max_quota,
+							} => {
+								ensure!(
+									!blocks_count.is_zero() && !quota_increment.is_zero() && !max_quota.is_zero(),
+									Error::<T>::InvalidRateLimitRule
+								);
+							}
+							_ => {}
```
xlc, line 205: better to explicitly list all the variants

xtokens/src/lib.rs
```diff
@@ -527,6 +536,23 @@ pub mod module {
...
+				// per asset check
+				let amount = match asset.fun {
+					Fungibility::Fungible(amount) => amount,
+					Fungibility::NonFungible(_) => 1,
+				};
+
+				let rate_limiter_id = T::RateLimiterId::get();
+
+				// check if the asset transfer from `who` can bypass the rate limiter.
+				if !T::RateLimiter::bypass_limit(rate_limiter_id, who.clone()) {
```
xlc, line 549:
```suggestion
				if !T::RateLimiter::bypass_limit(rate_limiter_id, &who) {
```

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+
+							*last_updated = now;
+							*remainer_quota = quota_increment
+								.saturating_mul(inc_times)
+								.saturating_add(*remainer_quota)
+								.min(max_quota);
+						}
+					}
+
+					_ => {}
```
shaunxw, line 355:
```suggestion
					Unlimited | NotAllowed  => {}
```

Will be good to explicitly list all cases.

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+
+		fn is_whitelist(limiter_id: Self::RateLimiterId, key: impl Encode) -> bool {
+			let encode_key: Vec<u8> = key.encode();
+
+			for key_filter in LimitWhitelist::<T>::get(limiter_id).0 {
+				match key_filter {
+					KeyFilter::Match(bounded_vec) => {
+						if encode_key == bounded_vec.into_inner() {
+							return true;
+						}
```
shaunxw, line 374:
```suggestion
						return encode_key == bounded_vec.into_inner();
```

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+					}
+					KeyFilter::StartsWith(prefix) => {
+						if encode_key.starts_with(&prefix) {
+							return true;
+						}
```
shaunxw, line 379:
```suggestion
						return encode_key.starts_with(&prefix);
```

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+					}
+					KeyFilter::EndsWith(postfix) => {
+						if encode_key.ends_with(&postfix) {
+							return true;
+						}
```
shaunxw, line 384:
```suggestion
						return encode_key.ends_with(&postfix);
```

rate-limit/src/lib.rs
wangjj9219: return false will break the for loop.

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+		Match(BoundedVec<u8, ConstU32<MAX_FILTER_KEY_LENGTH>>),
+		/// If the encoded key starts with the vec, the key is in whitelist.
+		StartsWith(BoundedVec<u8, ConstU32<MAX_FILTER_KEY_LENGTH>>),
+		/// If the encoded key ends with the vec, the key is in whitelist.
+		EndsWith(BoundedVec<u8, ConstU32<MAX_FILTER_KEY_LENGTH>>),
+	}
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
```
xlc, line 79: should also have a BlockNumberProvider so we can for example use relaychain block number if we want to

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+pub mod module {
+	use super::*;
+
+	/// Limit rules type.
+	#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo)]
+	pub enum RateLimitRule {
+		/// Each `blocks_count` blocks to reset remainer quota to `quota`
+		/// amount. is_allowed check return true when the remainer quota gte the
+		/// consume amount.
+		PerBlocks { blocks_count: u64, quota: u128 },
```
xlc, line 44: maybe we could have a `enum Period { Blocks(u64), Seconds(u64) }` and able to merge `PerBlocks` and `PerSeconds` into `PerPeriod { period: Period, quota: u128 }`

zjb0807 (approved):
LGTM

rate-limit/src/lib.rs
```diff
@@ -0,0 +1,454 @@
...
+
+	/// Match rules to fitler key is in bypass whitelist.
+	#[derive(PartialOrd, Ord, PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+	pub enum KeyFilter {
+		/// If the encoded key is equal to the vec, the key is in whitelist.
+		Match(BoundedVec<u8, ConstU32<MAX_FILTER_KEY_LENGTH>>),
+		/// If the encoded key starts with the vec, the key is in whitelist.
+		StartsWith(BoundedVec<u8, ConstU32<MAX_FILTER_KEY_LENGTH>>),
+		/// If the encoded key ends with the vec, the key is in whitelist.
+		EndsWith(BoundedVec<u8, ConstU32<MAX_FILTER_KEY_LENGTH>>),
```
xlc, line 78: I don't think we need this

xtokens/src/lib.rs
```diff
@@ -552,6 +561,18 @@ pub mod module {
...
+				// per asset check
+				let amount = match asset.fun {
+					Fungibility::Fungible(amount) => amount,
+					Fungibility::NonFungible(_) => 1,
+				};
+
+				let rate_limiter_id = T::RateLimiterId::get();
+
+				// try consume quota of the rate limiter.
+				T::RateLimiter::try_consume(rate_limiter_id, asset.id, amount, Some(who.clone()))
```
xlc, line 574:
```suggestion
				T::RateLimiter::try_consume(rate_limiter_id, asset.id, amount, Some(&who))
```

I think this should work?

xlc (commented):
there are some missing coverages reported by code coverage, but 100% sure if the report is correct but need to double check it

rate-limit/src/lib.rs
wangjj9219: If the key is AccountId, `Match` pattern may usually be used, but `StartWith` pattern can still be used to match all the pallet accounts. Can consider retaining it, and should use `StartWith` and `EndWith` with caution in actual configuration.

xlc (approved):

PR #857, opened 2022-11-29
Changed `vested_transfer` extrinsic behavior.

PraetorP:
Fixed self-vesting case (`from` == `to`) when it was possible to self-freeze funds above the current account balance.


xlc:
Please add some tests.

xlc (approved):

PR #859, opened 2022-12-07
feat: MinXcmFee from storage

sander2:
Allows you to store the MinXcmFee in storage. The advantage of this is that no runtime upgrade is required when you want to update this value. Usage for existing consumers is unchanged, other than having to provide `type WeightInfo = ();` now. To opt in into fetching the value from storage, users can use `type MinXcmFee = XTokens;`.

### DIFF
--- a/asset-registry/src/mock/para.rs
+++ b/asset-registry/src/mock/para.rs
@@ -336,6 +336,7 @@ impl orml_xtokens::Config for Runtime {
 	type LocationInverter = LocationInverter<Ancestry>;
 	type MaxAssetsForTransfer = MaxAssetsForTransfer;
 	type ReserveProvider = RelativeReserveProvider;
+	type WeightInfo = ();
 }
 
 impl orml_xcm::Config for Runtime {
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -24,6 +24,8 @@
 #![allow(clippy::boxed_local)]
 #![allow(clippy::too_many_arguments)]
 
+mod weights;
+
 use frame_support::{
 	log,
 	pallet_prelude::*,
@@ -42,6 +44,8 @@ use xcm::{latest::Weight, prelude::*};
 use xcm_executor::traits::{InvertLocation, WeightBounds};
 
 pub use module::*;
+pub use weights::WeightInfo;
+
 use orml_traits::{
 	location::{Parse, Reserve},
 	GetByKey, XcmTransfer,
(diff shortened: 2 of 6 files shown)


xlc (changes requested):
why do you need to make a new trait?

It is possible to achieve your goal without any code changes here. Checkout what we did at Acala
https://github.com/AcalaNetwork/Acala/blob/783feaa7094f4ce16acd04e5313f7c3977d8ba6e/runtime/karura/src/xcm_config.rs#L313
https://github.com/AcalaNetwork/Acala/blob/783feaa7094f4ce16acd04e5313f7c3977d8ba6e/modules/xcm-interface/src/lib.rs

we could have a new pallet in ORML to handle XCM fee configuration
our xcm-interface pallet include bunch Acala specific code but it is possible to spend bit more time to make it generic and resuable

sander2:
> why do you need to make a new trait?

It's not absolutely necessary, but the alternative is to implement `GetByKey` on the `Pallet<T>` which felt not so clean to me. But I'm happy to change that if you prefer.

> It is possible to achieve your goal without any code changes here. Checkout what we did at Acala https://github.com/AcalaNetwork/Acala/blob/783feaa7094f4ce16acd04e5313f7c3977d8ba6e/runtime/karura/src/xcm_config.rs#L313 https://github.com/AcalaNetwork/Acala/blob/783feaa7094f4ce16acd04e5313f7c3977d8ba6e/modules/xcm-interface/src/lib.rs
> 
> we could have a new pallet in ORML to handle XCM fee configuration our xcm-interface pallet include bunch Acala specific code but it is possible to spend bit more time to make it generic and resuable

I realize that there are workarounds possible, like putting the storage in other pallets. However, to me it made most sense to incorporate this into the xtokens pallet itself - I didn't want to add another pallet just for this one storage item. Having it inside XTokens itself is cleaner imo - it's easier for people to opt into it and it keeps everything a bit more tidy

xlc:
> It's not absolutely necessary, but the alternative is to implement GetByKey on the Pallet<T> which felt not so clean to me. But I'm happy to change that if you prefer.

We usually have a new struct and have `GetByKey` implemented on the new struct

xlc:
I think a generic XCM fee pallet will make sense that is able to store fee for various different operations. In future, I do believe parachains will want to support different kinds of xcm operations and that usually requires different amount of fee & weights

sander2:
> I think a generic XCM fee pallet will make sense that is able to store fee for various different operations. In future, I do believe parachains will want to support different kinds of xcm operations and that usually requires different amount of fee & weights

Could you clarify what you mean with different kinds of xcm operations? Do you mean messages that are being sent by pallets other than `xtokens`?

What kind of interface would you like to see on a generic XCM fee pallet?

xlc:
Yes other than xtokens. For example, we could have the [orml-xcm](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xcm/src/lib.rs) implements the method to open/accept hrmp channel.

The interface could be something similar to [xcm-interface](https://github.com/AcalaNetwork/Acala/blob/783feaa7094f4ce16acd04e5313f7c3977d8ba6e/modules/xcm-interface/src/lib.rs), just generalize it to be able to define the `XcmInterfaceOperation` in the runtime

PR #860, opened 2022-12-08
Upgrade polkadot-v0.9.34

wangjj9219:
need to replace the rev of cumulus after the cumulus polkadot-v0.9.34 released


shekohex:
Closes #862

Issue #861, opened 2022-12-12
Use explicit call indices


xlc:
https://github.com/paritytech/substrate/pull/12891

Issue #862, opened 2022-12-13
Bump to polkadot=0.9.34


tbraun96:
Polkadot will need to be bumped to 0.9.34 in order to be consonant with the emerging ecosystem.

PR #864, opened 2022-12-20
Upgrade to polkadot v0.9.36

CertainLach:


tokens/src/impls.rs
```diff
@@ -67,6 +67,10 @@ where
 			B::can_withdraw(asset, who, amount)
 		}
 	}
+
+	fn asset_exists(_asset: Self::AssetId) -> bool {
+		true
```
zjb0807, line 72:
```suggestion
		TestKey::contains(&asset)
```

CertainLach:
Closes #861 #862

tokens/src/lib.rs
```diff
@@ -1744,6 +1749,10 @@ impl<T: Config> fungibles::Inspect<T::AccountId> for Pallet<T> {
 	) -> WithdrawConsequence<Self::Balance> {
 		Self::withdraw_consequence(who, asset_id, amount, &Self::accounts(who, asset_id))
 	}
+
+	fn asset_exists(_asset: Self::AssetId) -> bool {
+		true
```
zjb0807, line 1754: @xlc Do you want to return false here? 
Prevent the substrate from using it later

tokens/src/impls.rs
CertainLach: Every asset that is not in `TestKey` (`B`), is delegated to `A`, I think that this implementation will be not consistent with other methods?

tokens/src/lib.rs
zjb0807: https://github.com/paritytech/substrate/pull/12782

`unimplemented!()` is fine.

zjb0807, line 1754:
```suggestion
		unimplemented!()
```

tokens/src/impls.rs
zjb0807: I think `unimplemented!()` is better

CertainLach: I propose returning false for now, just to be on the safe side, it would be bad if it caused a runtime panic.

tokens/src/impls.rs
```diff
@@ -67,6 +67,10 @@ where
 			B::can_withdraw(asset, who, amount)
 		}
 	}
+
+	fn asset_exists(_asset: Self::AssetId) -> bool {
+		false
```
xlc, line 72:
```
               if TestKey::contains(&asset) {
			true
		} else {
 			B::asset_exists(asset)
 		}
```

tokens/src/lib.rs
```diff
@@ -1744,6 +1749,10 @@ impl<T: Config> fungibles::Inspect<T::AccountId> for Pallet<T> {
 	) -> WithdrawConsequence<Self::Balance> {
 		Self::withdraw_consequence(who, asset_id, amount, &Self::accounts(who, asset_id))
 	}
+
+	fn asset_exists(_asset: Self::AssetId) -> bool {
+		false
```
xlc, line 1754: asset exists if TotalIssuance exists

xlc:
some tests for the new `asset_exists` implementation will be nice

PR #893, opened 2023-03-13
fix migration and tests

wangjj9219:
The key's concat type of the map storage `assetRegistry.locationToAssetId` is `Twox64Concat`:
https://github.com/open-web3-stack/open-runtime-module-library/blob/a730f7901844a2216ed773769ef69c08c742c2da/asset-registry/src/lib.rs#L100

In the previous migration test cases, using `Blake2_128Concat` as the concat type for the old key:
https://github.com/open-web3-stack/open-runtime-module-library/blob/a730f7901844a2216ed773769ef69c08c742c2da/asset-registry/src/migrations.rs#L30

https://github.com/open-web3-stack/open-runtime-module-library/blob/a730f7901844a2216ed773769ef69c08c742c2da/asset-registry/src/tests.rs#L613-L619

This seems masked some issues: 
there are many Multilocation with GeneralKey on Karura, some data lost during the migration tests on forked Karura. 

After debugging, it was found that atually if the Iterator here is not drained first, the insertion in the following for loop maybe cause skip some data:
https://github.com/open-web3-stack/open-runtime-module-library/blob/a730f7901844a2216ed773769ef69c08c742c2da/asset-registry/src/migrations.rs#L29-L37

### DIFF
--- a/asset-registry/src/migrations.rs
+++ b/asset-registry/src/migrations.rs
@@ -27,8 +27,9 @@ mod v2 {
 
 		weight.saturating_accrue(T::DbWeight::get().reads(1));
 		let old_data =
-			storage_key_iter::<xcm::v2::MultiLocation, T::AssetId, Blake2_128Concat>(module_prefix, storage_prefix)
-				.drain();
+			storage_key_iter::<xcm::v2::MultiLocation, T::AssetId, Twox64Concat>(module_prefix, storage_prefix)
+				.drain()
+				.collect::<sp_std::vec::Vec<_>>();
 
 		for (old_key, value) in old_data {
 			weight.saturating_accrue(T::DbWeight::get().writes(1));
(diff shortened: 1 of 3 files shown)


wangjj9219:
@tgmichel Hi, buddy, you should include this fix.

xlc:
We should prefer use [`translate`](https://github.com/paritytech/substrate/blob/ee316317b85b2f65fc022b27bbfefcd42b6560ae/frame/support/src/storage/mod.rs#L102) if possible.

Also a sidenote, read everything in memory is actually ok because the 5MB PoV limit will kick in way before the actual memory limit.

tgmichel:
> you should include this fix

Thank you @wangjj9219 

> prefer use [translate](https://github.com/paritytech/substrate/blob/ee316317b85b2f65fc022b27bbfefcd42b6560ae/frame/support/src/storage/mod.rs#L102) if possible

translate can be used to migrate values, not keys unfortunately (afaik)

xlc:
Right. It is not possible to do in place translate of the key while keeping the same storage name.

xlc (approved):

Issue #894, opened 2023-03-15
Polkadot-0.9.38 asset registry migration incorrect


sander2:
The asset registry contains a `VersionedMultiLocation` here: https://github.com/open-web3-stack/open-runtime-module-library/blob/8301b6c2871ce4e0279c78b3714e8e3a40a749ea/traits/src/asset_registry.rs#L29

This type changed in the 0.9.38 upgrade. Here's the old version:

https://github.com/paritytech/polkadot/blob/645723987cf9662244be8faf4e9b63e8b9a1b3a3/xcm/src/lib.rs#L73-L76

And here's the new: 
https://github.com/paritytech/polkadot/blob/8deef133d3ca1bdea8c6267c4a66ecabb903a18b/xcm/src/lib.rs#L318-L325

So assetRegistry.metadata needs to be migrated as well

xlc:
no add deps to 0.9.37 won't wrok. the only way is copy the definition

xlc:
The upside is that V1 should be identical to V2 (you should have some tests to check that). So if you don't already have V0 in the storage, you may be able to skip the migration. Otherwise you could upgrade it to V1 before the runtime upgrade.

sander2:
> The upside is that V1 should be identical to V2 (you should have some tests to confirm that). So if you don't already have V0 in the storage, you may be able to skip the migration. Otherwise you could upgrade it to V1 before the runtime upgrade.

V2 multilocation is indeed identical to V1 - it's just an import in 0.9.37: https://github.com/paritytech/polkadot/blob/645723987cf9662244be8faf4e9b63e8b9a1b3a3/xcm/src/v2/mod.rs#L68

A migration would still be required, but it'd be a simple one: `OldVersionMultiLocation::V1(x) --> VersionedMultiLocation::V2(x)`. Without a migration it'd get interpreted as `V3`.

Anyway, we don't have V0 ourselves, so that part of the migration is not needed for us, but I don't know if this is true for everyone. If you're comfortable with merging a migration that only works for `V1`, that'd work for me

xlc:
I think we all want to avoid doing a lot of work that may not be needed at all. So let's just do the migration for V1. We can make sure there are enough docs and warnings so the developer using the migration know they need to check the onchain data first. For example, the migration can be named like `MigrateV1Only_V0NotSupported`

sander2:
Addressed by https://github.com/paritytech/polkadot/pull/6884, which was backported to 0.9.38

PR #896, opened 2023-03-15
fix: asset-registry metadata migration

sander2:
Addresses https://github.com/open-web3-stack/open-runtime-module-library/issues/894 .

Note that as discussed in the issue, the migration only works if `Metadata.location` is `None` or `Some(VersionedMultiLocation::V1(_))`. Any metadata items that are `Some(VersionedMultiLocation::V0())` will be set to `None`.

### DIFF
--- a/asset-registry/src/lib.rs
+++ b/asset-registry/src/lib.rs
@@ -29,7 +29,7 @@ mod mock;
 mod tests;
 
 mod migrations;
-pub use migrations::Migration;
+pub use migrations::MigrateV1Only_V0NotSupported;
 
 #[frame_support::pallet]
 pub mod module {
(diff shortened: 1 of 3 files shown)


xlc:
Kind of blocked by https://github.com/paritytech/polkadot/pull/6884 which changes the encoding format for versioned types.

xlc:
Actually there is no need of migration for V1 because V1 and V2 are identical (after https://github.com/paritytech/polkadot/pull/6884 is merged)

xlc:
I am not sure if that makes sense. They should just upgrade it to the v1 format before upgrade to v3. Delete data and have invalid data is somewhat the same to the runtime as failed decode should just give the default data.

PR #900, opened 2023-04-03
remove require_transactional

zjb0807:
To support eth_call RPC `xtokens: Transfer failed: Transactional(NoLayer)`
https://github.com/AcalaNetwork/Acala/blob/54feaa65d729ea1e4f269d52fd4cbb2d0e304a48/modules/evm/rpc/src/lib.rs#L170

### DIFF
--- a/xtokens/src/lib.rs
+++ b/xtokens/src/lib.rs
@@ -35,7 +35,6 @@
 use frame_support::{
 	log,
 	pallet_prelude::*,
-	require_transactional,
 	traits::{Contains, Get},
 	Parameter,
 };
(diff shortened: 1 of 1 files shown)


xlc:
maybe that means you missing `transactional` somewhere? you can always wrap the runtime API call

zjb0807:
> maybe that means you missing `transactional` somewhere? you can always wrap the runtime API call

https://github.com/AcalaNetwork/Acala/commit/e2aae53156d6005b2c21c6f79040720608ea166f

image

It doesn't work

zjb0807:
I made a mistake, try again

Issue #919, opened 2023-05-30
DataFeeder: Allow pass the root account.


lemunozm:
Hi!

I need to insert values into the oracle from another pallet. I do not want to attach that pallet in a tight way, so I would like to use the `DataFeeder` to do that. Nevertheless, I cannot know exactly which account is part of the Oracle pallet's member list from another pallet perspective, and I can not use Root.

It would be great to have some `DataFeeder` method that does not need to use an account. Or at least, allow using an `Option<AccountId>` like the `do_feed_values()` in order to use Root.

If you agree with this modification, I could open a PR with this.

ermalkaleci:
That other pallet account_id can be another member maybe?

lemunozm:
The other pallet uses `DataFeeder` in order to create an initial worst state for benchmarking. Ideally, it could be great if a do not need to modify other pallets to set up an internal benchmark behavior.

xlc:
You could consider use the all zero account?

>  feed_value() only allow feeding one key/value from one account

I don't think this is true? One account can feed multiple keys?

I think we can add a new method to `DataFeeder` trait

`fn feed_values(who: Option<AccountId>, values: &[(Key, Value)] -> DispatchResult;`

and keep the existing one and provide a default implementation to call this new one

In this way we can minimize the breaking change

lemunozm:
> You could consider use the all zero account?

Not sure if I get what you mean here, could you expand? 🙏🏻 

> I don't think this is true? One account can feed multiple keys?

Looking at the signature seems like several calls to this works. Nevertheless, the Oracle implementation does not allow reusing the same account to fetch twice in the same block (getting a `AlreadyFeeded` error), so now I'm stuck fetching one value from one account.

> `fn feed_values(who: Option<AccountId>, values: &[(Key, Value)] -> DispatchResult;`

This is just what I was looking for, it would be awesome to have it! Although I think you can not create a default implementation for it, so it will be a breaking change.

lemunozm:
Another way to not add a new method breaking changes is that the oracle's implementation of `feed_value` skips the `HasDispatched` part.

I mean, the `HasDispatched` check makes sense for external interaction with the oracle, but it seems like a significant restriction from an internal usage using the trait

xlc:
All zero account is basically `AccountId::default()` or `0x0000...0000`

[...]

PR #920, opened 2023-05-31
Oracle: allow multiple feed values from inner calls

lemunozm:
# Description

This PR shows an example of an implementation to allow calling `feed_value()` multiple times internally, avoiding the `AlreadyFeeded` error without being a breaking change and without effecting to the extrinsic behavior

Related issue: #919

### DIFF
--- a/traits/src/data_provider.rs
+++ b/traits/src/data_provider.rs
@@ -2,9 +2,9 @@ use sp_runtime::DispatchResult;
 use sp_std::vec::Vec;
 
 /// Data provider with ability to provide data with no-op, and provide all data.
-pub trait DataFeeder<Key, Value, AccountId>: DataProvider<Key, Value> {
+pub trait DataFeeder<Key, Value, AccountId> {
 	/// Provide a new value for a given key from an operator
-	fn feed_value(who: AccountId, key: Key, value: Value) -> DispatchResult;
+	fn feed_value(who: Option<AccountId>, key: Key, value: Value) -> DispatchResult;
 }
 
 /// A simple trait to provide data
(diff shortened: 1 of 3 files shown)


xlc:
I think this works. Just need some tests and docs.

lemunozm:
Would you agree to add to this PR a minor breaking change to allow passing `Option<AccountId>` to `feed_value()`? Would be much more ergonomic to have a "default" member easy to use.

xlc:
I think it is ok.
It is a breaking change which isn't ideal, but otherwise the code itself won't be ideal as well. It is not worth to sacrifice code maintainability for such trivial breaking change.

traits/src/data_provider.rs
```diff
@@ -2,9 +2,9 @@ use sp_runtime::DispatchResult;
 use sp_std::vec::Vec;
 
 /// Data provider with ability to provide data with no-op, and provide all data.
-pub trait DataFeeder<Key, Value, AccountId>: DataProvider<Key, Value> {
+pub trait DataFeeder<Key, Value, AccountId> {
```
lemunozm, line 5: I think not having the dependency on `DataProvider` allows entities to implement only the feeder part, which can be helpful. This should not be an API change, but please feel free to tell me if you prefer not, and I revert it.

lemunozm (commented):
I think this is ready for review. I've also added the test part related to these changes.

traits/src/data_provider.rs
xlc: that's fine

xlc (approved):

PR #934, opened 2023-07-12
Fix MessageQueue pallet order in mocks

Agusrodri:
This change will prevent getting some encoding errors with XCM related tests in the future.

Edit: To give more context, the reason why this change is needed is because the new `MessageQueue` pallet should be placed at the same index where the old `ParasUmp` pallet was situated before it was removed in the upgrade to v0.9.43. 

If we test the execution of an upward message with the current mock configuration, in most of the cases it will pass without any problem, but in some other ones it will fail while processing the XCM Transact instruction, more specifically while decoding the call that this instruction receives. This was the case in Moonbeam for instance, where we encountered that some XCM tests were failing due to this while testing the upgrade to v0.9.43. These issues got solved by placing the new `MessageQueue` pallet in the correct index.

### DIFF
--- a/asset-registry/src/mock/relay.rs
+++ b/asset-registry/src/mock/relay.rs
@@ -222,7 +222,7 @@ construct_runtime!(
 		System: frame_system::{Pallet, Call, Storage, Config, Event<T>},
 		Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>},
 		ParasOrigin: origin::{Pallet, Origin},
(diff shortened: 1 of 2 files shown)


xlc:
Can you provide elaborate bit more why this is needed? If the orders are important, we need some code comments to explicitly document the requirement.

Agusrodri:
> Can you provide elaborate bit more why this is needed?

Yes of course, I've just updated the PR's description.

xlc (approved):

Issue #947, opened 2023-08-25
Vesting to move tokens to account derived from receiver and vesting to allow gov delegation


dzmitry-lahoda:
it is expected that team vested people are bening and should be able to use future tokens to vote.

so allow to setup proxy for vested account as desired by host runtime config to that account.

which will be usual voting

xlc:
I do not understand the request. Vested token can be used for voting. There is no need to have proxy.

dzmitry-lahoda:
https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/conviction-voting/src/lib.rs#L625

PR #948, opened 2023-08-26
vesting: can be used for gov

dzmitry-lahoda:
for https://github.com/open-web3-stack/open-runtime-module-library/issues/947


xlc:
I am not sure if we need to emphasize this. If you want to add something, it should be saying the vesting implementation is use locks to allow the tokens to be used by other pallets that's also using locks, for example, the democracy pallet.

PR #961, opened 2023-10-25
use crates.io

xlc:
blocked by https://github.com/paritytech/polkadot-sdk/pull/2017

### DIFF
--- a/tokens/src/mock.rs
+++ b/tokens/src/mock.rs
@@ -5,7 +5,10 @@
 use super::*;
 use frame_support::{
 	construct_runtime, parameter_types,
-	traits::{ChangeMembers, ConstU32, ConstU64, ContainsLengthBound, Everything, SortedMembers},
+	traits::{
+		tokens::{PayFromAccount, UnityAssetBalanceConversion},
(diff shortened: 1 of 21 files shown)


tokens/src/mock.rs
```diff
@@ -117,15 +121,23 @@ impl pallet_treasury::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type OnSlash = ();
 	type ProposalBond = ProposalBond;
-	type ProposalBondMinimum = ProposalBondMinimum;
-	type ProposalBondMaximum = ProposalBondMaximum;
-	type SpendPeriod = SpendPeriod;
```
zjb0807, line 122: remove the unused `ProposalBondMinimum`, `ProposalBondMaximum`, `SpendPeriod`, `MaxApprovals`

zjb0807 (commented):
The versions of the pallet are different. Will it be more troublesome to upgrade in the future? Check it one by one?

xlc:
https://github.com/paritytech/diener/pull/42 should solve the problem.

I am currently using https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates to ensure they are all latest version

PR #964, opened 2023-11-08
Check Rust feature configs

ggwpez:
Adds the [Zepter](https://github.com/ggwpez/zepter) CLI to the CI to verify feature integrity of all Rust crates. It uses (nearly) the same config file as the [Polkadot-SDK](https://github.com/paritytech/polkadot-sdk/blob/640e385aec02ee8b80284febe679ce785ade66b9/.config/zepter.yaml).  
cc @AurevoirXavier maybe you also want to add your tool?  

What it does:
- Formats the features with 80 line width.
- Checks that `try-runtime`, `runtime-benchmarks` and `std` is forwarded to dependencies.  
- Skips the check that if crate `B` has a feature that  crate `A` also must have that feature (we skipped this in the SDK, but could be enabled).


AurevoirXavier:
> Adds the [Zepter](https://github.com/ggwpez/zepter) CLI to the CI to verify feature integrity of all Rust crates. It uses (nearly) the same config file as the [Polkadot-SDK](https://github.com/paritytech/polkadot-sdk/blob/640e385aec02ee8b80284febe679ce785ade66b9/.config/zepter.yaml). cc @AurevoirXavier maybe you also want to add your tool?
> 
> What it does:
> 
> * Formats the features with 80 line width.
> * Checks that `try-runtime`, `runtime-benchmarks` and `std` is forwarded to dependencies.
> * Skips the check that if crate `B` has a feature that  crate `A` also must have that feature (we skipped this in the SDK, but could be enabled).

Sure, thank you. However, I am currently waiting for my new laptop, so I will be unavailable for a brief period.

ggwpez:
> Sure, thank you. However, I am currently waiting for my new laptop, so I will be unavailable for a brief period.

Okay nice. I am waiting for @xlc for some feedback on the approach.

Makefile
```diff
@@ -34,9 +34,11 @@ Cargo.toml: Cargo.dev.toml
 
 dev-format: Cargo.toml
 	cargo fmt --all
+	zepter format features --fix
```
xlc, line 37: maybe have them on a separate target otherwise we need to indicate on README to tell user to install `zepter` first

but a good amount of time people won't need to use it so I want to make it optional

.github/workflows/test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
       uses: dtolnay/rust-toolchain@nightly
     - name: Install Wasm toolchain
       run: rustup target add wasm32-unknown-unknown
+    - name: Install Zepter
+      run: cargo install zepter --version 0.14.0 --locked -q -f --no-default-features && zepter --version
```
xlc, line 33: doesn't need to happen in this PR but at some stage I want to use https://github.com/taiki-e/install-action (powered by https://github.com/cargo-bins/cargo-binstall) to install tools. I don't know if zepter is compatible with it but we could try and see

.github/workflows/test.yml
```diff
@@ -41,3 +43,5 @@ jobs:
       run: make dev-check
     - name: Run tests
       run: make dev-test
+    - name: Check Rust features
+      run: zepter run check
```
xlc, line 47: best to do check before tests

Makefile
ggwpez: Okay I put it into another target. LMK if i overdid it 😅

.github/workflows/test.yml
ggwpez: Yea i wanted to use `cargo-install` as well. I think it just pulls binaries from the the GH release page. Will explore it and then open another MR when its done.

.github/workflows/test.yml
ggwpez: Done

xlc (approved):

asset-registry/Cargo.toml
```diff
@@ -57,34 +57,68 @@ orml-xcm-support = { path = "../xcm-support", default-features = false }
...
-	"polkadot-runtime-common/runtime-benchmarks",
 ]
 try-runtime = [
+	"cumulus-pallet-dmp-queue/try-runtime",
+	"cumulus-pallet-parachain-system/try-runtime",
+	"cumulus-pallet-xcm/try-runtime",
+	"cumulus-pallet-xcmp-queue/try-runtime",
 	"frame-support/try-runtime",
 	"frame-system/try-runtime",
+	"orml-tokens/try-runtime",
```
xlc, line 114: this is causing issues when I try to publish this to crates.io using `cargo release`. orml-tokens is only a dev-dependencies so I think there is no reason to enable features for it?

ggwpez: Hm, maybe it should be an optional feature enable then like `orml-tokens?/try-runtime`.  
I will try to debug it.

ggwpez: No the optional is not possible to compile then. I think in the past, Parity stripped dev-dependencies before publishing from the `Cargo.toml` files. Not sure if this still happens.

xlc: I tried and not working. I believe it is `cargo release` removes dev-dependencies before publish and all those features will not be valid.
Besides, we don't really need to enable them for dev-dependencies anyway.

ggwpez: There is a first version of this here https://github.com/ggwpez/zepter/pull/56 for `zepter transpose features strip-dev-only`.  
I will write some more tests tomorrow and fix a few edge-cases but it seems to work when i tested it with `cargo publish --dry-run` on the crates here.

AurevoirXavier:
cargo-featalign requires a `kitchensink-runtime` or similar to perform the checks which is more accurate.

PR #966, opened 2023-11-16
[XNFT PoC] xtokens non-fungible multiassets support

mrshiposha:
A change to the `do_transfer_multiassets`: support of the `NonFungible` fungibility for non-fee assets.

This PR doesn't change the API of extrinsics that use the configurable `CurrencyId` since it is a PoC. In the future, we could alter the `amount` argument to take the fungibility into account.

So only the following extrinsics can be used to transfer an NFT asset:
* `transferMultiasset`
* `transferMultiassetWithFee`
* `transferMultiassets`


xlc (changes requested):
need tests

xlc (commented):
some tests to ensure sending nft are actually working?

xlc (approved):

PR #967, opened 2023-11-16
Remove dev-only features

ggwpez:
Ignores features of dependencies that are only used for `dev` dependencies. I think this can cause build error, but lets see.  
So maybe we need to add the features directly into the deps. Downside is that this will cause build errors when the tests are run without that feature. 

Another way is to just call the strip command as part of the release process and never push the result of that. I think this would be the most correct way.


xlc:
Ideally, tests shouldn't require additional feature flag. They should be unconditionally enabled in dev-dependencies. There could be exceptions such as some features could be mutual exclusive (which is a code smoke) but we don't have those atm.

xlc (approved):

PR #977, opened 2024-02-13
deps: update substrate-sdk to version 1.7.0

RomarQ:
Update polkadot-sdk to version 1.7.0

Includes many xcm changes, which were introduced in https://github.com/paritytech/polkadot-sdk/pull/1230.

To reduce the amount of breaking changes I did not remove the `multi` prefix from extrinsics in the `xtokens` pallet:

```
`transfer_multiasset`: Transfer `MultiAsset` assets.
`transfer_multiasset_with_fee`: Transfer `MultiAsset` specifying the fee
`transfer_multicurrencies`: Transfer several currencies specifying the
`transfer_multiassets`: Transfer several `MultiAsset` specifying the item
```

I tested the changes against https://github.com/paritytech/polkadot-sdk/tree/release-polkadot-v1.7.0

### DIFF
--- a/asset-registry/src/lib.rs
+++ b/asset-registry/src/lib.rs
@@ -16,7 +16,7 @@ use sp_runtime::{
 use sp_std::prelude::*;
 use xcm::{v3::prelude::*, VersionedLocation};
 
-use impls::*;
+pub use impls::*;
 use module::*;
 pub use weights::WeightInfo;
 
--- a/asset-registry/src/lib.rs
+++ b/asset-registry/src/lib.rs
@@ -14,7 +14,7 @@ use sp_runtime::{
 	DispatchResult,
 };
 use sp_std::prelude::*;
-use xcm::{v3::prelude::*, VersionedMultiLocation};
+use xcm::{v4::prelude::*, VersionedLocation};
 
 pub use impls::*;
 pub use module::*;
@@ -67,7 +67,7 @@ pub mod module {
 	pub enum Error<T> {
 		/// Asset was not found.
 		AssetNotFound,
-		/// The version of the `VersionedMultiLocation` value used is not able
+		/// The version of the `VersionedLocation` value used is not able
 		/// to be interpreted.
 		BadVersion,
 		/// The asset id is invalid.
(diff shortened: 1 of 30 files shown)


RomarQ:
Updating dependencies in Cargo.dev.toml since the crates have been updated in crates.io.

RomarQ:
@xlc Can you have a look at the changes? Thank you in advance :)

asset-registry/src/lib.rs
```diff
@@ -14,7 +14,7 @@ use sp_runtime::{
 	DispatchResult,
 };
 use sp_std::prelude::*;
-use xcm::{v3::prelude::*, VersionedMultiLocation};
+use xcm::{v4::prelude::*, VersionedLocation};
```
xlc, line 17: please keep this v3, otherwise it is a breaking change that requires migration
while we can migrate to v4, let's wait until v4 is battle tested first and there is a clear deprecation plan for v3

xlc (changes requested):
I prefer to leave upgrade to XCM v4 on a separate PR and I prefer only upgrade to it when it is thoroughly tested (e.g. when it is used by system parachains on testnets)

RomarQ:
> I prefer to leave upgrade to XCM v4 on a separate PR and I prefer only upgrade to it when it is thoroughly tested (e.g. when it is used by system parachains on testnets). There were enough bugs on v3 upgrades and we shouldn't take lightly on upgrading to v4.

Not sure it will be possible to not upgrade to XCM v4 since `xcm-executor` uses `xcm::latest`, I will try to have something minimal with a converter between V4 and V3.

I think the only breaking change regarding the SCALE encoding is the `AssetId` since the `Abstract` has been removed.

xlc:
I will say it will be a major regression in polkadot-sdk if we are forced to use xcm v4 and if that’s actually the case, we need polkadot-sdk to be patched first.

asset-registry/src/lib.rs
RomarQ: The SCALE encoding did not change with the new xcm::v4::Location, I added a test to just to be sure https://github.com/open-web3-stack/open-runtime-module-library/pull/977/commits/38f45f2d9d7e652040444630f12679f99e8acbe0

I could add a migration, but seems unnecessary and expensive.

RomarQ: The same applies to pallet `unknown-tokens`, it does not require a migration.

xlc:
with https://github.com/paritytech/polkadot-sdk/pull/3328 we won't get 100% compatibility between v4 types and v3 types

wangjj9219 (approved):

zjb0807 (approved):

xlc (approved):

PR #979, opened 2024-02-23
Upgrade rust 1.75.0

zjb0807:
https://github.com/paritytech/srtool/pull/80


zjb0807:
https://crates.io/crates/staging-xcm-executor/versions

There is no cargo.lock, `staging-xcm-executor` is upgraded to 7.0.3

xlc (approved):
they are not allowed to make patch release with breaking change 💁‍♂️

PR #981, opened 2024-03-12
[WIP] xtokens integrate delay-tasks to delay transfer_assets

wangjj9219:


delay-tasks/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+			delay_task.on_cancel()?;
+			DelayedTaskQueue::<T>::remove(execute_block, id);
+
+			Self::deposit_event(Event::<T>::DelayedTaskCanceled { id });
+			Ok(())
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		fn _on_finalize(now: BlockNumberFor<T>) {
```
xlc, line 144: should do this with `on_idle` to be safer and we will know how many tasks to execute based on remaining weights

delay-tasks/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+pub use module::*;
+
+mod mock;
+mod tests;
+
+pub const DELAY_TASK_ID: [u8; 8] = *b"orml/dts";
+
+/// A delayed origin. Can only be dispatched via `dispatch_as` with a delay.
+#[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+pub struct DelayedExecuteOrigin;
```
xlc, line 35: should try to reuse the one from orml-authority. we can move it to orml-utils

xlc (commented):
a lot of feature are similar to orml-authoirty, maybe we can just reuse it? e.g. for schedule and cancel task

I see two ways:
modifty orml-authority to use this pallet to do the scheduling code
or have orml-auhtoirty to implement some trait and have this pallet use it

delay-tasks/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+			Self::deposit_event(Event::<T>::DelayedTaskAdded {
+				id,
+				task,
+				execute_block,
+			});
+			Ok(())
+		}
+	}
+
+	pub struct DelayedXtokensTaskHooks<T>(PhantomData<T>);
```
xlc, line 263: we should avoid xcm/xtokens code in this pallet

PR #990, opened 2024-04-10
improve self transfer handling

xlc:
improve self transfer handling for transfer_share_and_rewards

### DIFF
--- a/rewards/src/lib.rs
+++ b/rewards/src/lib.rs
@@ -329,11 +329,15 @@ impl<T: Config> Pallet<T> {
 		move_share: T::Share,
 		other: &T::AccountId,
 	) -> DispatchResult {
-		SharesAndWithdrawnRewards::<T>::mutate(pool, other, |increased_share| {
+		SharesAndWithdrawnRewards::<T>::try_mutate(pool, other, |increased_share| {
 			let (increased_share, increased_rewards) = increased_share;
-			SharesAndWithdrawnRewards::<T>::mutate_exists(pool, who, |share| {
+			SharesAndWithdrawnRewards::<T>::try_mutate_exists(pool, who, |share| {
 				let (share, rewards) = share.as_mut().ok_or(Error::<T>::ShareDoesNotExist)?;
 				ensure!(move_share < *share, Error::<T>::CanSplitOnlyLessThanShare);
+				if who == other {
+					// self transfer is noop
+					return Ok(());
+				}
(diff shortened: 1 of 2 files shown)


rewards/src/tests.rs
```diff
@@ -593,3 +593,35 @@ fn transfer_share_and_rewards() {
...
+		assert_noop!(
+			RewardsModule::transfer_share_and_rewards(&ALICE, &DOT_POOL, 200, &ALICE),
+			Error::<Runtime>::CanSplitOnlyLessThanShare
+		);
+
+		let pool_info = RewardsModule::pool_infos(DOT_POOL);
+		assert_ok!(RewardsModule::transfer_share_and_rewards(&ALICE, &DOT_POOL, 33, &ALICE));
+
+		let new_pool_info = RewardsModule::pool_infos(DOT_POOL);
+		assert_eq!(pool_info, new_pool_info, "reward transfer does not affect the pool");
```
zjb0807, line 616: Should it be transferred to BOB to verify this case?

rewards/src/lib.rs
```diff
@@ -329,11 +329,15 @@ impl<T: Config> Pallet<T> {
 		move_share: T::Share,
 		other: &T::AccountId,
 	) -> DispatchResult {
-		SharesAndWithdrawnRewards::<T>::mutate(pool, other, |increased_share| {
+		SharesAndWithdrawnRewards::<T>::try_mutate(pool, other, |increased_share| {
 			let (increased_share, increased_rewards) = increased_share;
-			SharesAndWithdrawnRewards::<T>::mutate_exists(pool, who, |share| {
+			SharesAndWithdrawnRewards::<T>::try_mutate_exists(pool, who, |share| {
 				let (share, rewards) = share.as_mut().ok_or(Error::<T>::ShareDoesNotExist)?;
 				ensure!(move_share < *share, Error::<T>::CanSplitOnlyLessThanShare);
```
ermalkaleci, line 336: should we add like ED for shares? It can be a const with min possible amount so transfer or withdrawal won't leave dust

xlc: yeah in next PR

rewards/src/tests.rs
xlc: this is to test transfer to self, which previously wasn't handled correctly

zjb0807 (approved):

PR #1003, opened 2024-05-30
polkadot v1.12.0

clangenb:
## Noteworthy changes 
* Update benchmarking macro. I essentially integrated the changes from: https://github.com/paritytech/polkadot-sdk/pull/3934/.
* Add `Preservation` arg to burn function in fungible trait
* Add `XcmRecorder` to `XcmConfig` for mocks

I would be very happy if this is released soon on crates.io ;)

### DIFF
--- a/Cargo.dev.toml
+++ b/Cargo.dev.toml
@@ -37,35 +37,35 @@ scale-info = { version = "2.10.0", default-features = false, features = ["derive
 serde = { version = "1.0.189" }
 parity-scale-codec = { version = "3.6.5", default-features = false, features = ["max-encoded-len"] }
 
-cumulus-pallet-xcm = { version = "0.11.0", default-features = false }
-cumulus-primitives-core = { version = "0.11.0", default-features = false }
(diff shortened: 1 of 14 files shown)


zjb0807:
Please update https://github.com/open-web3-stack/open-runtime-module-library/blob/master/Cargo.dev.toml to use polkadot-sdk v1.12.0

clangenb:
I see that the coverage workflow failed due to the rate limit. Can we merge it like this?

zjb0807:
> I could have noticed this earlier if the CI ran. I am wondering: as this isn't my first contribution, and neither my last. Is it possible to make the CI run with my PRs automatically?

This should be a GitHub setting. But it seems to run automatically.

> I see that the coverage workflow failed due to the rate limit. Can we merge it like this?

The coverage is not required. It's fine.

clangenb:
Alright I see. I thought that I read that it needs a maintainer to run the action.

Cool, thanks a lot!

xlc (approved):

